PR #11, opened 2020-12-30
Remove unused lock.

mcdee:
This removes a mutex lock followed by an unlock that has no code
inbetween.  This was a remnant from when job management was handled
within the control function, and is no longer required.


gary-rowe (approved):

TezBaker (commented):
Looks good

PR #12, opened 2020-12-30
Max attestation delay

mcdee:
This adds a flag `controller.max-attestation-delay` that controls the maximum time that Vouch will wait until it starts to generates attestations for a slot.


gary-rowe (approved):
Looks good to me

TezBaker (commented):
Looks good

PR #13, opened 2021-01-31
Handle chain reorganisations.

mcdee:
When a chain reorganisation occurs it is possible that this impacts Vouch's attestations and proposals for the current and next epoch.  This patch listens to the Ethereum 2 events stream for reorganisation notifications.  If it finds one it cancels existing and creates new duties, as required.


gary-rowe (approved):

TezBaker (commented):
Checked over the logic for cancelling and rescheduling the tasks and makes sense

PR #57, opened 2022-02-02
Add target scoring for block proposals.

mcdee:
Scoring target votes requires historical information for chain, however it is not feasible to fetch this at the time the block is proposed. This fetches blocks as they are added to the chain, and uses the information to score target votes accordingly.


StevenBerryman (commented):
There was a deepsource failure with this change

PR #130, opened 2023-03-04
Fix dirk default timeout bug

dB2510:
This PR fixes dirk default timeout bug. It used to set default timeout of 30s for dirk account manager irrespective of the provided account manager. This led to vouch expecting dirk account manager in `startAccountManager` function even if wallet account manager is provided. Refer:
https://github.com/attestantio/vouch/blob/2bbfe67138f56b45beb6bdd5219caa833316e4f7/main.go#L233
https://github.com/attestantio/vouch/blob/2bbfe67138f56b45beb6bdd5219caa833316e4f7/main.go#L931

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,12 +86,12 @@ import (
 	firstsynccommitteecontributionstrategy "github.com/attestantio/vouch/strategies/synccommitteecontribution/first"
 	"github.com/attestantio/vouch/util"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	homedir "github.com/mitchellh/go-homedir"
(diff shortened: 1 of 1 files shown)


mcdee:
Thank you for reporting this issue.  I'd rather not split out the defaults, and so have taken a different approach to this issue in #131 which should also be more robust in terms of deciding when to start which account manager.  Please could you take a look at this and let me know if you are happy with that alternative patch?

dB2510:
Hey @mcdee 
#131 looks good to me. Closing this one in the favour of #131.
Thanks!

PR #131, opened 2023-03-06
Tidy up accountmanager startup.

mcdee:
Accountmanager was starting Dirk on an incorrect basis, and multiple accountmangers could be defined and fail to start silently.  Address both of these issues.


dB2510 (approved):
LGTM!
Thanks Jim!

StevenBerryman (approved):

PR #179, opened 2024-02-09
Eliminate duplicates from pubKeys

catwith1hat:
This is one potential fix for the duplicate key issue.

One thing that I am not entirely sure about is that I am opening multiple wallets concurrently with:
```
s.openWallet(ctx, walletName)
```
I have too little insights into the concurrency requirements around wallet management to understand if thats safe. Before this commit, everything was done serially. Please review.

### DIFF
--- a/services/accountmanager/dirk/service.go
+++ b/services/accountmanager/dirk/service.go
@@ -162,7 +162,6 @@ func (s *Service) refreshAccounts(ctx context.Context) {
 	defer span.End()
 
 	// Create the relevant wallets.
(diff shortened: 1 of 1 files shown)


mcdee:
Thank you for this patch.  Regarding your question:

> One thing that I am not entirely sure about is that I am opening multiple wallets concurrently...

`openWallet()` locks a mutex for its entire duration, so it's okay for this to be called by multiple goroutines simultaneously.

mcdee (approved):

catwith1hat:
Thanks for the review and quick merge!

PR #206, opened 2024-05-13
feat: add blockrelay.priority-builders

zimbatm:
Priority builders are a list of public keys of builders from which bids will be accepted first, even if the bid is lower. If no bids are received from the priority builders, bids from other builders will be accepted.

This is useful when you have a special contract with specific relays and you would like to have their blocks used instead of other relays. While still having other relays as a fallback.

I wanted to submit this to you early to see if this is something you would be interested in having.


mcdee (changes requested):
Thank you for this PR.  I have made a couple of immediate comments on it.

I'm wondering if it would be better to have additional fields in the existing `blockauctioneer.Results` rather than creating a second instance of the structure, as that would keep a single struct being returned and would lead to being more extensible in future, but I'm not convinced.  Thoughts?

mcdee:
I think it makes sense to merge this pretty much as-is, we can look at a more generic way of handling this later on if needs be.

One item: please could you fix the commit to the `docs/configuration.md` so that it only includes the required changes rather than the reformat?  Thanks.

zimbatm:
Great, thanks for reviewing our work @mcdee

MrKoberman:
Made a small change to pass the deepsource check.

mcdee (approved):
Looks good, and separating out the code that sets the builder bid makes this a lot cleaner than the original version.  Nice work.

PR #212, opened 2024-07-10
Verify sync committee participation on head event

Bez625:
- Add VerifySyncCommitteeMessages to the head event handler for tracing verification of sync committee participation
- Add recording of sync committee messages to the sync committee messenger


services/controller/standard/events.go
```diff
@@ -332,3 +337,41 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+
+// VerifySyncCommitteeMessages handles the "head" events from the beacon node for sync committee validation.
+func (s *Service) VerifySyncCommitteeMessages(ctx context.Context, data *apiv1.HeadEvent) {
+	_, span := otel.Tracer("attestantio.vouch.services.controller.standard").Start(ctx, "VerifySyncCommitteeMessages")
+	defer span.End()
+
+	logger := log.With().Uint64("slot", uint64(data.Slot)).Logger()
+	logger.Trace().Msg("Received head event")
+
+	rootReported, found := s.syncCommitteeMessenger.GetBeaconBlockRootReported(data.Slot)
```
mcdee, line 349: I don't think that this is the correct logic.  In slot `x` the sync committee message is generated using the head of the chain, which is usually the root of block `x`.  So in slot `x+1` (i.e. here) we want to look for the data for the previous slot, and compare it to the head block's parent hash.

services/controller/standard/events.go
```diff
@@ -332,3 +337,41 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+	if !found {
+		logger.Trace().Msg(fmt.Sprintf("no reported sync committee message data for slot: %d, skipping validation", data.Slot))
+		return
+	}
+
+	if rootReported.Root != data.Block {
+		logger.Trace().Msg(fmt.Sprintf("mismatch in block root for slot: %d. Reported: %s and observed: %s", data.Slot, rootReported.Root.String(), data.Block.String()))
+		// TODO: Check that it is correct to return here. If we do not match the block root we shouldn't be part of the SyncAggregate
```
mcdee, line 357: This is correct, you can return here.

services/controller/standard/events.go
```diff
@@ -332,3 +337,41 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+	if err != nil {
+		logger.Trace().Msg(fmt.Sprintf("failed to retrieve block: %s for slot: %d with err: %v", data.Block.String(), data.Slot, err))
+		return
+	}
+	syncAggregate, err := blockResponse.Data.SyncAggregate()
+	if err != nil {
+		logger.Trace().Msg(fmt.Sprintf("failed to get sync aggregate for block: %s for slot: %d with err: %v", data.Block.String(), data.Slot, err))
+		return
+	}
+	for _, index := range rootReported.ValidatorIndices {
```
mcdee, line 372: There is a mismatch here.  The sync committee bits are for the sync committee, not the validators themselves.  In other words `SyncCommitteeBits` is always 512 bits long, regardless of how many validators we have.

The sync committee duty provides the mapping from validator index to sync committee index.  This information is obtained elsewhere in the codebase, so may be able to be sneaked into existing structures rather than refetching it every time we are in this call.

mcdee (changes requested):
Some style tweaks and some more substantive logic issues.

services/controller/standard/events.go
```diff
@@ -343,35 +345,53 @@ func (s *Service) VerifySyncCommitteeMessages(ctx context.Context, data *apiv1.H
...
+
+	// Logging with the current slot as that is when this code is executed.
+	logger := log.With().Uint64("slot", uint64(currentSlot)).Logger()
 	logger.Trace().Msg("Received head event")
 
-	rootReported, found := s.syncCommitteeMessenger.GetBeaconBlockRootReported(data.Slot)
+	previousSlotData, found := s.syncCommitteeMessenger.GetDataUsedForSlot(previousSlot)
 	if !found {
-		logger.Trace().Msg(fmt.Sprintf("no reported sync committee message data for slot: %d, skipping validation", data.Slot))
+		logger.Trace().Msg(fmt.Sprintf("No reported sync committee message data for slot: %d, skipping validation", previousSlot))
```
mcdee, line 358: As we are using structured logging it makes sense to keep variables out of the message.  So, for example, this would look like:

```
logger.Trace().Uint64("slot",uint64(previousSlot)).Msg("No reported sync committee message data for slot; skipping validation")
```

Please update similarly throughout.

mcdee (changes requested):
Looking much better.  A couple of style items, and you'll need to fetch the parent root of the slot in the handler (probably best to use the client's `BeaconBlockHeader()` method for this).

services/controller/standard/events.go
```diff
@@ -332,3 +339,61 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+		Str("block_header", data.Block.String()).
+		Logger()
+	log.Trace().Msg("Received head event")
+
+	previousSlotData, found := s.syncCommitteeMessenger.GetDataUsedForSlot(previousSlot)
+	if !found {
+		log.Trace().Msg("No reported sync committee message data for slot; skipping validation")
+		return
+	}
+	blockResponse, err := s.signedBeaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
```
mcdee, line 365: Fetching the block is relatively heavyweight compared to just fetching the block header, which should be all that is needed here.

services/controller/standard/events.go
```diff
@@ -332,3 +339,61 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+
+	// We verify against the previous slot as that is when the sync committee will have reported.
+	previousSlot := data.Slot - 1
+	currentSlot := data.Slot
+
+	// Logging with the current slot as that is when this code is executed.
+	log := log.With().
+		Uint64("current_slot", uint64(currentSlot)).
+		Uint64("previous_slot", uint64(previousSlot)).
+		Str("block_header", data.Block.String()).
```
mcdee, line 356: This should be `block_root`, I believe.

services/controller/standard/events.go
```diff
@@ -332,3 +339,61 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+		Logger()
+	log.Trace().Msg("Received head event")
+
+	previousSlotData, found := s.syncCommitteeMessenger.GetDataUsedForSlot(previousSlot)
+	if !found {
+		log.Trace().Msg("No reported sync committee message data for slot; skipping validation")
```
mcdee, line 362: Any path that is unexpected (such as this one) but doesn't have an impact on the direct workings of the function we should log at `Debug` level.  This fits with the use of `Debug` level elsewhere.

This is true for a few other log messages in this function as well.

services/controller/standard/events.go
```diff
@@ -332,3 +339,61 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+	if err != nil {
+		log.Trace().Err(err).Msg("Failed to retrieve head block for sync committee validation")
+		return
+	}
+	parentRoot, err := blockResponse.Data.ParentRoot()
+	if err != nil {
+		log.Trace().Err(err).Msg("Failed to get parent root of head block")
+		return
+	}
+	if parentRoot.String() != previousSlotData.Root.String() {
```
mcdee, line 377: Rather than cast to string to compare, it would be cheaper to use `bytes.Equal(parentRoot[:],previousSlotData.Root[:])`

services/controller/standard/events.go
```diff
@@ -332,3 +339,61 @@ func (s *Service) subscribeToBeaconCommittees(ctx context.Context,
...
+		log.Trace().Str("head_parent_root", parentRoot.String()).Str("broadcast_root", previousSlotData.Root.String()).
```
mcdee, line 378: Rather than `Str("head_parent_root", parentRoot.String())`  you can use `Stringer("head_parent_root", parentRoot)` which calls the `String()` function internally.

mcdee (changes requested):

services/controller/standard/events.go
Bez625: I understand the point and have addressed for other lines, but I think we _do_ expect this log line frequently. That is because the slot data is only stored when we broadcast a sync committee message, so most of the time we would not find previous slot data.

services/controller/standard/events.go
Bez625: I had a quick look at the `beaconBlockHeadersProvider` and couldn't see any way to get the `SyncAggregate` data for us to check if our validators were included. Is there something I'm missing or a lighter way to get the `SyncAggregate` data?

services/controller/standard/events.go
mcdee: Ah right, so ideally we would want to only run this function if we knew we were participating in a sync committee, but that's going to be slightly harder on the logic side.  Let's leave this as-is for now, then.

services/controller/standard/events.go
mcdee: Yep my bad I didn't look past the use of the block for the parent root.

PR #215, opened 2024-07-16
Updates for 1.8.2

mcdee:
Fixes for interoperability with lodestar 1.20.


services/beaconblockproposer/standard/propose.go
```diff
@@ -334,13 +236,10 @@ func (*Service) confirmProposalData(_ context.Context,
...
 	}
 
-	proposalGraffiti, err := proposal.Graffiti()
-	if err != nil {
-		return errors.Wrap(err, "failed to obtain proposal graffiti")
-	}
-	if !bytes.Equal(proposalGraffiti[:], graffiti[:]) {
-		return errors.New("proposal data contains incorrect graffiti")
-	}
+	// RANDAO reveal can be different in DVT situations, so do not check it.  It wil have already been checked by the underlying
```
Bez625, line 239: Small typo

```suggestion
	// RANDAO reveal can be different in DVT situations, so do not check it. It will have already been checked by the underlying
```

strategies/beaconblockproposal/best/score.go
```diff
@@ -14,680 +14,30 @@
...
-	"github.com/attestantio/go-eth2-client/spec/deneb"
-	"github.com/attestantio/go-eth2-client/spec/phase0"
-	"github.com/prysmaticlabs/go-bitfield"
 )
 
 // scoreBeaconBlockPropsal generates a score for a beacon block.
-// The score is relative to the reward expected by proposing the block.
-func (s *Service) scoreBeaconBlockProposal(ctx context.Context,
+// The score is the reward expected by proposing the block.
+func (*Service) scoreBeaconBlockProposal(_ context.Context,
```
Bez625, line 25: Unless I'm missing something I think this is only called in one place so we can probably remove the unused context argument.

strategies/beaconblockproposal/best/score.go
```diff
@@ -14,680 +14,30 @@
...
-	"github.com/attestantio/go-eth2-client/spec"
-	"github.com/attestantio/go-eth2-client/spec/altair"
-	"github.com/attestantio/go-eth2-client/spec/bellatrix"
-	"github.com/attestantio/go-eth2-client/spec/capella"
-	"github.com/attestantio/go-eth2-client/spec/deneb"
-	"github.com/attestantio/go-eth2-client/spec/phase0"
-	"github.com/prysmaticlabs/go-bitfield"
 )
 
 // scoreBeaconBlockPropsal generates a score for a beacon block.
```
Bez625, line 23: Typo in comment.

```suggestion
// scoreBeaconBlockProposal generates a score for a beacon block.
```

Bez625 (changes requested):
Some minor comments and it looks like the linters failed. I had a quick look and seems to be a couple of magic numbers detected in the updated eth2client mock.

strategies/beaconblockproposal/best/score.go
mcdee: Because this is a backport of existing code we want to keep this for now.  It can be revisited in master.

mcdee:
Linting will fail due to this being a backport, and we aren't going to backport all of the lint configuration changes as well as we want to move to the 1.9 releases soon.

PR #223, opened 2024-08-01
Update builder client to 0.5.0

mcdee:


strategies/builderbid/best/builderbid.go
```diff
@@ -431,10 +432,15 @@ func (*Service) obtainBid(ctx context.Context,
 ) {
 	log := zerolog.Ctx(ctx).With().Str("bidder", provider.Address()).Logger()
 
-	builderBid, err := provider.BuilderBid(ctx, slot, parentHash, pubkey)
+	builderBidResponse, err := provider.BuilderBid(ctx, &builderapi.BuilderBidOpts{
```
Bez625, line 435: FYI `builderBidResponse` shadows the struct name on line 45. I'm not sure this has a material difference on the code change so will approve, but thought it worth mentioning.

Bez625 (approved):

strategies/builderbid/best/builderbid.go
mcdee: Yeah that's worth addressing.  I've pushed an update to avoid this issue.

Bez625 (approved):

PR #226, opened 2024-08-02
Add deadline builder bid strategy.

mcdee:


mcdee:
Initial commit of deadline builder bid.  Introduction of excluded and privileged relay configuration means that the code is a little more convoluted than I would like; a future commit will integrate these items into a single structure and reduce code complexity as a result.

Bez625 (approved):
Added a couple of small suggestions. I've approved in case you want to dismiss them.

PR #227, opened 2024-08-07
Builder configurations

mcdee:
Unify "privileged" and "excluded" builders, and generalise into the idea of  per-builder configurations.  This change significantly simplifies the auction bidding code.


Bez625 (changes requested):
Maybe i've missed something, but there seems to be a lot of passing around an unused context. I've suggested changes to remove them, apologies if these were put in intentionally and I've missed the reason why.

mcdee:
Although context may not be used in certain places immediately it's very handy to keep it around.  There are a few areas where it can be very useful:

- adding in tracing, for example if a particular function is suspected of running slowly
- adding contextual logging
- allowing updates to functions to allow calls to external entities (commonly for timeouts, for example adding in caching to an existing call)

Although it is possible to add context in at the point we need it, passing it down through functions from the off makes subsequent patches much smaller and clearer.  Also, if the function is public then it means a change to the published API and that's not a good thing if we can avoid it, so we try to put context everywhere and just ignore it when it isn't being used by _ing the variable.

Bez625 (approved):

PR #233, opened 2024-08-13
feat: changing builder configs factor to be a percentage

MrKoberman:
Boost is a percentage multiplier and should accommodate for the same requirements as factor but also gives the validator more precise control on how much they are willing to increase privileged relays score.

One additional feature of this is that validators could also create a new group and reduce the score of a specific relay. At the moment we don't see this as an issue, but would love to get your opinion on this.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -221,19 +221,19 @@ blockrelay:
   log-results: true
   # builder-configs contain specific configurations for different builders, with each builder defined by its public key.
   # The base score for each bid is the value to the proposer, in wei.  The final score is calculated by adding
-  # the 'offset' value for the specific builder, and then multiplying it by the 'factor' value.  For example,
-  # if the base value is 1000, the offset is 10 and the factor is 2 then the final score is (1000+10)*2 = 2020.  If the
-  # offset is not configured it defaults to 0; if the factor is not configured it defaults to 1.  The category is used
+  # the 'offset' value for the specific builder, and then multiplying by the percentage of 'boost' value. For example,
+  # if the base value is 1000, the offset is 10 and the boost is 110 then the final score is (1000+10)*(110/100) = 1111.  If the
+  # offset is not configured it defaults to 0; if the boost is not configured it defaults to 100.  The category is used
   # for differentiation of bids in metrics.
   builder-configs:
     '0xaaaa...':
       category: 'privileged'
-      # With a factor of 1000000000 bids from this builder are pretty much guaranteed to be included above bids from other builders.
-      factor: 1000000000
+      # With boost of 1000000000 bids from this builder are pretty much guaranteed to be included above bids from other builders.
+      boost: 1000000000
     '0xbbbb...':
       category: 'excluded'
       # With a factor of 0 bids from this builder will be ignored.
-      factor: 0
+      boost: 0
 
 # tracing sends OTLP trace data to the supplied endpoint.
 tracing:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -221,19 +221,19 @@ blockrelay:
   log-results: true
   # builder-configs contain specific configurations for different builders, with each builder defined by its public key.
   # The base score for each bid is the value to the proposer, in wei.  The final score is calculated by adding
-  # the 'offset' value for the specific builder, and then multiplying by the percentage of 'boost' value. For example,
-  # if the base value is 1000, the offset is 10 and the boost is 110 then the final score is (1000+10)*(110/100) = 1111.  If the
-  # offset is not configured it defaults to 0; if the boost is not configured it defaults to 100.  The category is used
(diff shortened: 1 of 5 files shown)


Bez625:
Thanks for the PR @MrKoberman. We took a look through it and request that you revert to using the name `factor` and making it a percentage. This will reduce the amount of changes required in the PR.

mcdee:
@MrKoberman we're looking at releasing a beta in the next few days, at which point we'd want to freeze the features.  As such, if you could make the changes as mentioned above by @Bez625 (leave the name as factor, make it a percentage, stick with integer maths) ASAP it will allow us to include this.  Once the feature is in a release it will be much harder for us to change.

strategies/builderbid/best/builderbid.go
```diff
@@ -243,7 +243,12 @@ func (*Service) setBuilderBid(ctx context.Context,
 		score = new(big.Int).Add(score, builderConfig.Offset)
 	}
 	if builderConfig.Factor != nil {
-		score = new(big.Int).Mul(score, builderConfig.Factor)
+		if builderConfig.Factor.Cmp(big.NewInt(0)) > 0 {
```
mcdee, line 246: This seems overly complex.  What about `score = new(big.Int).Div(new(big.Int).Mul(score, builderConfig.Factor), big.NewInt(100))`?

Note that if we want to disallow factors <0 then we should do so when reading in the configuration rather than here.

strategies/builderbid/deadline/builderbid.go
```diff
@@ -308,7 +308,12 @@ func (*Service) setBuilderBid(ctx context.Context,
 		score = new(big.Int).Add(score, builderConfig.Offset)
 	}
 	if builderConfig.Factor != nil {
-		score = new(big.Int).Mul(score, builderConfig.Factor)
+		if builderConfig.Factor.Cmp(big.NewInt(0)) > 0 {
```
mcdee, line 311: Same comment as above.

mcdee (changes requested):
Some comments on the method.

strategies/builderbid/best/builderbid.go
mcdee: (Note that another PR has addressed that factors cannot be <0)

MrKoberman:
Thank you for the review, @mcdee, @Bez625. I addressed the comments on the PR with the latest commit.

docs/configuration.md
```diff
@@ -221,14 +221,14 @@ blockrelay:
   log-results: true
   # builder-configs contain specific configurations for different builders, with each builder defined by its public key.
   # The base score for each bid is the value to the proposer, in wei.  The final score is calculated by adding
-  # the 'offset' value for the specific builder, and then multiplying it by the 'factor' value.  For example,
-  # if the base value is 1000, the offset is 10 and the factor is 2 then the final score is (1000+10)*2 = 2020.  If the
-  # offset is not configured it defaults to 0; if the factor is not configured it defaults to 1.  The category is used
+  # the 'offset' value for the specific builder, and then multiplying by the percentage of 'factor' value. For example,
+  # if the base value is 1000, the offset is 10 and the factor is 110 then the final score is ((1000+10)/110)*100 = 1100.  If the
```
mcdee, line 225: This is sitll incorrect.  It should read `(1000+10)*110/100 = 1111`

mcdee (changes requested):
One documentation change, the rest looks okay.

docs/configuration.md
MrKoberman: Yes but the code will not produce `1111` but `1100` 

```
score = new(big.Int).Div(new(big.Int).Mul(score, builderConfig.Factor), big.NewInt(100))
```

docs/configuration.md
```diff
@@ -221,14 +221,14 @@ blockrelay:
   log-results: true
   # builder-configs contain specific configurations for different builders, with each builder defined by its public key.
   # The base score for each bid is the value to the proposer, in wei.  The final score is calculated by adding
-  # the 'offset' value for the specific builder, and then multiplying it by the 'factor' value.  For example,
-  # if the base value is 1000, the offset is 10 and the factor is 2 then the final score is (1000+10)*2 = 2020.  If the
-  # offset is not configured it defaults to 0; if the factor is not configured it defaults to 1.  The category is used
+  # the 'offset' value for the specific builder, and then multiplying by the percentage of 'factor' value. For example,
+  # if the base value is 1000, the offset is 10 and the factor is 110 then the final score is ((1000+10)/110)*100 = 1111.  If the
```
mcdee, line 225: This calculation is still incorrect.  It should be `(1000+10)*110/100`

mcdee (changes requested):

docs/configuration.md
MrKoberman: Sorry did not notice this. Fixed

mcdee (approved):

PR #234, opened 2024-08-13
Tidy up sync committee logs

Bez625:


mcdee:
Due to the way that sync committee aggregates work (broken out into 4 chunks) it would be good to have committee index information as well.  So, for example, for included you would have `includedValidatorIndices` and `includedCommitteeIndices` and you'd append both values for each included sync committee message.  That would allow us to spot if all of the messages that weren't included were in a single aggregate, for example.

mcdee (approved):

PR #239, opened 2024-08-22
Reduce bare interfaces and type assertions

Bez625:
Hey @mcdee - I'd like to get your feedback on this. Essentially what I'd like to do is reduce the number of type assertions and bare interface{} parameters. The overall approach is to use closures with input data of different types enclosed so that we can use the generic job scheduler without having to cast from bare interfaces in the specific jobFunc. 

I appreciate this is changing existing interfaces and I'm not really sure the impact of that outside of vouch. I do think it removes a lot of code we should never enter in to at runtime and allows for specific types in the appropriate interfaces/implementations.


mcdee:
At first glance this looks like a good move.  Can this be extended so that the current `ScheduleJobNoData` is rolled back into `ScheduleJob`?  I know that this would break interfaces, but if it results in a cleaner result I'm okay with that.

Bez625:
> At first glance this looks like a good move. Can this be extended so that the current `ScheduleJobNoData` is rolled back into `ScheduleJob`? I know that this would break interfaces, but if it results in a cleaner result I'm okay with that.

Yes and I think we have two options:

* Replace `ScheduleJob` with `ScheduleJobNoData` - I don't believe there are usages of `ScheduleJob` in this PR (outside of tests) so it should just be a case of replacing. I can then have a look at the other `Schedule*` functions to see if similar changes can be made. 
* Use `ScheduleJob` "as is", but call it with nil data and have an unused data argument in the closure, e.g.:

```golang
s.scheduler.ScheduleJob(ctx,
    "Aggregate attestations",
    fmt.Sprintf("Beacon block attestation aggregation for slot %d committee %d", attestation.Data.Slot, attestation.Data.Index),
    s.chainTimeService.StartOfSlot(attestation.Data.Slot).Add(s.attestationAggregationDelay),
    func(ctx context.Context, _ any) { s.attestationAggregator.Aggregate(ctx, aggregatorDuty) },
    nil,
)
```

My preference would be the former as it enforces the behaviour, has far fewer unused parameters and doesn't require implicit knowledge of the pattern/workaround on the call side. The latter would at least leave the scheduler interface intact, but I personally think it's worth taking the hit if we're going to change other interfaces.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 dev:
   - change default timestamp in logs to millisecond-precision
   - allow custom timestamp formatting in logs
+  - refactor scheduler and associated interfaces
```
mcdee, line 4: We generally don't add non-user-actionable changes in the changelog, so this can be removed.

services/controller/standard/service.go
```diff
@@ -369,9 +369,11 @@ func (s *Service) epochTicker(ctx context.Context, data interface{}) {
 		"Epoch",
 		fmt.Sprintf("Prepare for epoch %d", currentEpoch+1),
 		s.chainTimeService.StartOfEpoch(currentEpoch).Add(time.Duration(offset)*time.Second),
-		s.prepareForEpoch,
-		&prepareForEpochData{
-			epoch: currentEpoch + 1,
+		func(ctx context.Context) {
+			prepData := &prepareForEpochData{
```
mcdee, line 373: Do you need to assign this here, or can you just pass the struct directly (i.e. `s.prepareForEpoch(ctx,&prepareForEpochData{...`)?

services/scheduler/service.go
```diff
@@ -22,6 +22,9 @@ import (
 // JobFunc is the type for jobs.
 type JobFunc func(context.Context, interface{})
 
+// JobFuncNoData is the type for jobs.
+type JobFuncNoData func(context.Context)
```
mcdee, line 26: Do we still need `JobFuncNoData`, or can we make this `JobFunc` and remove the existing definition?

mcdee (changes requested):
Mainly looks good; a few questions/requests.

services/scheduler/service.go
Bez625: `JobFunc` is needed for the periodic schedule method. I have a PR ready to make similar changes to the periodic schedule method and that will also remove `JobFuncNoData`.

mcdee (approved):

PR #247, opened 2024-08-29
Disable sync committee fast-track by default

Bez625:
Hi @mcdee I wanted to get your thoughts on this change. I feel if we leave fast-track on by default it could significantly change the behaviour of sync committee participation when upgrading from 1.8.2 to 1.9.0 without being obvious to the users why. 

I'd lean towards this being more of an explicit "opt-in" change rather than enabled by default, hence the change.


mcdee:
We already carry out fast track sync committees by default, so the only difference should be the use of the strategies to obtain data.  Although this should increase sync committee message inclusion, it doesn't make a fundamental difference to how the system works so I don't think that it makes sense to turn it off by default.

PR #251, opened 2024-09-13
Use the multi signer logic for sync committees

Bez625:


services/synccommitteemessenger/standard/service.go
```diff
@@ -176,47 +176,44 @@ func (s *Service) Message(ctx context.Context, duty *synccommitteemessenger.Duty
...
+		accounts[i] = account
+	}
+	sigs, err := s.contributions(ctx, accounts, s.chainTimeService.SlotToEpoch(duty.Slot()), *beaconBlockRoot)
+	if err != nil {
+		s.log.Error().Err(err).Msg("Failed to sign sync committee messages")
+		return nil, errors.Wrap(err, "failed to sign sync committee messages")
+	}
+
+	for i := range accounts {
+		if len(sigs) == 0 || len(sigs) < i {
```
mcdee, line 200: It looks like this length check doesn't need to be in the loop, it could sit after fetching `sigs` and be executed only once.

services/signer/standard/helpers.go
```diff
@@ -60,3 +60,52 @@ func (*Service) sign(ctx context.Context,
...
+		container := phase0.SigningData{
+			ObjectRoot: root,
+			Domain:     domain,
+		}
+		root, err := container.HashTreeRoot()
+		if err != nil {
+			return []phase0.BLSSignature{}, errors.Wrap(err, "failed to generate hash tree root")
+		}
+		for i := range accounts {
+			sig, err := accounts[i].(e2wtypes.AccountSigner).Sign(ctx, root[:])
```
mcdee, line 103: It would be good to do a type check for this rather than just casting it.

main.go
```diff
@@ -110,7 +110,7 @@ import (
 )
 
 // ReleaseVersion is the release version for the code.
-var ReleaseVersion = "1.9.0-beta.3"
+var ReleaseVersion = "1.9.0-beta.dev"
```
mcdee, line 113: Please change this to 1.10.0-dev

mcdee (changes requested):
A few changes required, plus looks like it needs a rebase.

services/synccommitteemessenger/standard/service.go
Bez625: `sigs` is a fixed size array so this check didn't really make sense anyway. I have replaced with a zero check on the sig for active accounts.

mcdee (approved):

PR #259, opened 2024-10-03
Sign slot selections with multi-sign

Bez625:


services/signer/standard/signslotselection.go
```diff
@@ -35,9 +35,6 @@ func (s *Service) SignSlotSelection(ctx context.Context,
 	ctx, span := otel.Tracer("attestantio.vouch.services.signer.standard").Start(ctx, "SignSlotSelection")
 	defer span.End()
 
-	var messageRoot phase0.Root
```
mcdee, line 38: Why has this been removed?

Is `SignSlotSelection` used any more?

mcdee (changes requested):

services/signer/standard/signslotselection.go
Bez625: I don't believe `messageRoot` was ever used. You can see further down we create another identical var named `slotBytes` that we eventually sign. 

That being said with these changes `SignSlotSelection` is no longer used by `vouch`. Neither are the below from the previous related PRs:
- `SignSyncCommitteeRoot`
- `SignSyncCommitteeSelection`
- `SignContributionAndProof`

Should I remove these methods and their implementations?

mcdee: Okay, great.

PR #272, opened 2024-10-30
Fix default gas limit on execution config

aivarasko:
Fix default gas limit on execution config

Example not working gas limit on execution config - instead of 40000000, relay would use default 3000000.

```
{
  "version": 2,
  "fee_recipient": "...redacted...",
  "relays": {
    "https://holesky.titanrelay.xyz": {
      "public_key": "0xaa58208899c6105603b74396734a6263cc7d947f444f396a90f7b7d3e65d102aec7e5e5291b27e08d02c50a050825c2f"
    }
  },
  "proposers": [],
  "gas_limit": "40000000"
}
```


mcdee (approved):

mcdee:
Thanks for the fix!

PR #280, opened 2025-01-03
Add versioned attestation submission and various electra updates

Bez625:
Hey Jim, this is a very rough initial draft for an electra branch. I thought it better to get your input at this stage before I continued. 

The main change here is adding logic to handle the new attestation data format. The branch depends on https://github.com/attestantio/go-eth2-client/pull/178. 

I've not looked at the following yet as I thought I better get some feedback on the approach:
- Units for the versioned attestation submission.
- Aggregates and proofs for the versioned attestations.


services/attester/standard/attest.go
```diff
@@ -140,19 +153,28 @@ func (s *Service) attest(
...
 	s.log.Trace().Dur("elapsed", time.Since(started)).Msg("Signed")
+	var attestations []*spec.VersionedAttestation
 
-	attestations := s.createAttestations(ctx, duty, accounts, committeeIndices, validatorCommitteeIndices, committeeSizes, data, sigs)
-	if len(attestations) == 0 {
-		s.log.Info().Msg("No signed attestations; not submitting")
-		return attestations, nil
+	if epoch < s.electraForkEpoch {
+		attestations = s.createAttestations(ctx, duty, accounts, committeeIndices, validatorCommitteeIndices, committeeSizes, data, sigs)
+		if len(attestations) == 0 {
```
mcdee, line 160: This check is duplicated, you could move it outside of the separate branches.

You may also want to consider making the ifs a switch, which is tidier and allows for easier expansion whenever the next change happens.

services/blockrelay/standard/auctionblock.go
```diff
@@ -103,17 +104,33 @@ func (s *Service) cacheBid(_ context.Context,
...
 		// No bid supplied; create a dummy for the purposes of caching so that if asked for this bid
 		// we can actively respond to say we don't have anything (as opposed to attempting to fetch
 		// a bid when we're queried).
 		s.log.Trace().Msg("Bid is nil; creating dummy")
-		bid = &builderspec.VersionedSignedBuilderBid{
-			Version: spec.DataVersionDeneb,
-			Deneb: &deneb.SignedBuilderBid{
-				Message: &deneb.BuilderBid{
-					Value: uint256.NewInt(0),
+		// It would be nice if we could avoid doing this as it would remove the need to set the electraForkEpoch on the service.
```
mcdee, line 113: Do we care about the fork version of this bid?  From what I remember (and this is not looking at the code right now) we only use this dummy bid to call `Value()`, so we can get away with always returning with the deneb version specified and populated, with no ill effects.

services/controller/standard/attester.go
```diff
@@ -143,7 +143,8 @@ func (s *Service) AttestAndScheduleAggregate(ctx context.Context, duty *attester
 	}
 	log.Trace().Dur("elapsed", time.Since(started)).Msg("Attested")
 
-	if len(attestations) == 0 || attestations[0].Data == nil {
+	firstAttestationData, err := attestations[0].Data()
```
mcdee, line 146: This will crash if len(attestations) == 0 so we need to check that first.

services/controller/standard/service.go
```diff
@@ -558,6 +563,33 @@ func (s *Service) handleBellatrixForkEpoch(ctx context.Context) {
 	}()
 }
 
+// fetchElectraForkEpoch fetches the epoch for the electra hard fork.
```
mcdee, line 566: Should consider replacing this with the function added in chaintime to obtain the hard fork, although I recognise that it's a copy of the other similar functions already in the controller.

services/submitter/immediate/service.go
```diff
@@ -106,16 +106,16 @@ func (s *Service) SubmitProposal(ctx context.Context, proposal *api.VersionedSig
 }
 
 // SubmitAttestations submits multiple attestations.
-func (s *Service) SubmitAttestations(ctx context.Context, attestations []*phase0.Attestation) error {
+func (s *Service) SubmitAttestations(ctx context.Context, opts *api.SubmitAttestationsOpts) error {
 	ctx, span := otel.Tracer("attestantio.vouch.services.submitter.immediate").Start(ctx, "SubmitAttestations")
 	defer span.End()
 
-	if len(attestations) == 0 {
```
mcdee, line 113: Should check that opts != nil first, to avoid a potential crash.

services/submitter/multinode/submitattestations.go
```diff
@@ -31,13 +32,13 @@ import (
...
 
 // SubmitAttestations submits a batch of attestations.
-func (s *Service) SubmitAttestations(ctx context.Context, attestations []*phase0.Attestation) error {
+func (s *Service) SubmitAttestations(ctx context.Context, opts *api.SubmitAttestationsOpts) error {
 	ctx, span := otel.Tracer("attestantio.vouch.service.submitter.multinode").Start(ctx, "SubmitAttestations", trace.WithAttributes(
 		attribute.String("strategy", "multinode"),
 	))
 	defer span.End()
 
-	if len(attestations) == 0 {
```
mcdee, line 40: Should check that opts != nil first, to avoid a potential crash.

mcdee (changes requested):
A few tweaks suggested.  Note that we will want to run this on both electra and deneb testnets before merging, to ensure we're happy with it.

mcdee (changes requested):
A few tweaks suggested.  Note that we will want to run this on both electra and deneb testnets before merging, to ensure we're happy with it.

Bez625:
> A few tweaks suggested. Note that we will want to run this on both electra and deneb testnets before merging, to ensure we're happy with it.

Thanks Jim, the only concern I'd have with running on deneb is that the beacon clients would have to handle the new v2 endpoint for submitting attestations due to the changes we made to https://github.com/attestantio/go-eth2-client/blob/electra/http/submitattestations.go#L50. 

I had an earlier version that did continue to submit to the v1 endpoint if pre-electra so that we could handle existing clients, but if you prefer us to keep at v2 we can keep these changes off master and do some ad hoc tests with new tests.

mcdee:
Yeah, we would have to make a note of the minimum consensus client versions for support, but they should all have support in for this endpoint by now.  We can test on mekong to confirm.

Bez625:
> Yeah, we would have to make a note of the minimum consensus client versions for support, but they should all have support in for this endpoint by now. We can test on mekong to confirm.

Understood. I have marked this as draft to prevent it getting merged in to master so it can be the feature branch for pectra.

PR #288, opened 2025-02-05
Electra fix index

Bez625:


services/attester/standard/attest.go
```diff
@@ -79,11 +79,13 @@ func (s *Service) Attest(ctx context.Context, duty *attester.Duty) ([]*spec.Vers
 	}
 	committeeIndices := make([]phase0.CommitteeIndex, len(validatingAccounts))
 	validatorCommitteeIndices := make([]phase0.ValidatorIndex, len(validatingAccounts))
+	validatorOrderedIndices := make([]phase0.ValidatorIndex, len(validatingAccounts))
```
mcdee, line 82: This appears to be a direct copy of `accountValidatorIndices`, can we not pass that instead?

mcdee (approved):

PR #294, opened 2025-02-10
bugfix - Attestation Data BlockRootToSlot return value

oleg-ssvlabs:
**Description**  
In cases where a slot does not exist in the cache (`map[phase0.Root]phase0.Slot` field), it is [fetched](https://github.com/attestantio/vouch/blob/master/services/cache/standard/blockroottoslot.go#L35) from the Consensus node. When successfully fetched, it should presumably be [returned](https://github.com/attestantio/vouch/blob/master/services/cache/standard/blockroottoslot.go#L47) by the function. However, instead of returning the newly fetched value, the function returns the value fetched from the cache, which means it will return `0`. This behavior appears to be a bug.  

This PR also introduces basic response validation for deeply nested pointer-type items to prevent potential panics

### DIFF
--- a/services/cache/standard/blockroottoslot.go
+++ b/services/cache/standard/blockroottoslot.go
@@ -17,6 +17,7 @@ import (
 	"context"
 
 	"github.com/attestantio/go-eth2-client/api"
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


Bez625:
Hi @oleg-ssvlabs - thank you very much for the contribution! We are close to the electra hardfork and so are focusing on releasing the changes necessary for that next. As a result we've included equivalent changes in to the feature branches for that release (we split the code between https://github.com/attestantio/vouch/pull/295 and https://github.com/attestantio/go-eth2-client/pull/199). I'd anticipate us releasing an electra compatible beta version of vouch in the next day or so.

Bez625:
Hi @oleg-ssvlabs - I'll close this now as the code has been added to the beta release for vouch v1.10

oleg-ssvlabs:
Sure, thank you @Bez625

PR #306, opened 2025-03-10
Always fetch execution configuration.

### LINKED ISSUE #305: Fetch the execution config even if you have only external validators
If there are no validator accounts handled by Vouch itself, it doesn't load the execution config: `"No validating accounts; not fetching execution config"`

However, we can have other clients and beacon nodes using vouch as the  external builder, even if Vouch itself doesn't do any attestation duties.
In this case, Vouch should fetch the execution config to be able to build blinded blocks with the correct relays, fee recipients, etc... for these accounts externally managed.

mcdee:
Fixes #305

Also update go version and related dependencies.


go.mod
```diff
@@ -1,60 +1,60 @@
...
 
 toolchain go1.23.2
 
 require (
 	github.com/attestantio/go-block-relay v0.4.3
-	github.com/attestantio/go-builder-client v0.6.0
-	github.com/attestantio/go-eth2-client v0.24.0
-	github.com/aws/aws-sdk-go v1.55.5
+	github.com/attestantio/go-builder-client v0.6.1
+	github.com/attestantio/go-eth2-client v0.24.1-0.20250226000601-fb893f25a47a
```
Bez625, line 10: Is the version of go-eth2-client intended? It looks like it's possibly a branch

Bez625 (commented):
Just one question on the version of go-eth2-client

go.mod
mcdee: Yep, that's not a good version.  I have pushed an update with a release version; thanks.

Bez625 (approved):

PR #312, opened 2025-04-02
Multi instance

mcdee:
Create a framework for multi instance Vouch, along with an initial implementation that uses attestation data to confirm if other Vouch instances are active or not.

### DIFF
--- a/services/multiinstance/always/metrics.go
+++ b/services/multiinstance/always/metrics.go
@@ -66,16 +66,11 @@ func registerPrometheusMetrics(_ context.Context) error {
 }
 
 // monitorActive provides metrics for the instance activity status.
-func monitorActive(operation string, active bool) {
+func monitorActive(operation string) {
 	if activeCounter == nil {
 		return
 	}
 
-	if active {
-		activeCounter.WithLabelValues(operation, "active").Inc()
-		activeState.Set(1)
-	} else {
-		activeCounter.WithLabelValues(operation, "inactive").Inc()
-		activeState.Set(0)
-	}
+	activeCounter.WithLabelValues(operation, "active").Inc()
+	activeState.Set(1)
 }
--- a/services/multiinstance/always/metrics.go
+++ b/services/multiinstance/always/metrics.go
@@ -1,0 +1,81 @@
+// Copyright © 2024 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package always
+
+import (
+	"context"
+
+	"github.com/attestantio/vouch/services/metrics"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	activeCounter *prometheus.CounterVec
+	activeState   prometheus.Gauge
+)
+
+func registerMetrics(ctx context.Context, monitor metrics.Service) error {
+	if activeCounter != nil {
+		// Already registered.
+		return nil
+	}
+	if monitor == nil {
+		// No monitor.
+		return nil
+	}
+	if monitor.Presenter() == "prometheus" {
+		return registerPrometheusMetrics(ctx)
+	}
+	return nil
+}
+
+func registerPrometheusMetrics(_ context.Context) error {
+	activeCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "vouch",
+		Subsystem: "multiinstance",
+		Name:      "checks_total",
+		Help:      "The number of checks for the active state",
+	}, []string{"operation", "result"})
+	if err := prometheus.Register(activeCounter); err != nil {
+		return errors.Wrap(err, "failed to register vouch_multiinstance_checks_total")
+	}
+
+	activeState = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "vouch",
+		Subsystem: "multiinstance",
+		Name:      "active",
+		Help:      "1 if this instance is active, otherwise 0",
+	})
+	if err := prometheus.Register(activeState); err != nil {
+		return errors.Wrap(err, "failed to register vouch_multiinstance_active")
+	}
+
+	return nil
+}
+
+// monitorActive provides metrics for the instance activity status.
+func monitorActive(operation string, active bool) {
+	if activeCounter == nil {
+		return
+	}
+
+	if active {
+		activeCounter.WithLabelValues(operation, "active").Inc()
+		activeState.Set(1)
+	} else {
+		activeCounter.WithLabelValues(operation, "inactive").Inc()
+		activeState.Set(0)
+	}
+}
--- a/services/multiinstance/staticdelay/metrics.go
+++ b/services/multiinstance/staticdelay/metrics.go
@@ -1,0 +1,81 @@
+// Copyright © 2024 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package staticdelay
+
+import (
+	"context"
+
+	"github.com/attestantio/vouch/services/metrics"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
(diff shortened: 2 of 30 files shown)


services/multiinstance/staticdelay/metrics.go
```diff
@@ -0,0 +1,81 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/always/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+func monitorActive(operation string, active bool) {
+	if activeCounter == nil {
+		return
+	}
+
+	if active {
+		activeCounter.WithLabelValues(operation, "active").Inc()
+		activeState.Set(1)
+	} else {
+		activeCounter.WithLabelValues(operation, "inactive").Inc()
```
Bez625, line 78: Given these are the `always` specific metrics I don't think we can ever enter this `else` block.

services/multiinstance/staticdelay/onattestationfailure.go
```diff
@@ -0,0 +1,27 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/onproposalfailure.go
```diff
@@ -0,0 +1,28 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/parameters.go
```diff
@@ -0,0 +1,127 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/service.go
```diff
@@ -0,0 +1,103 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/service_test.go
```diff
@@ -0,0 +1,162 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/shouldattest.go
```diff
@@ -0,0 +1,73 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/shouldattest.go
```diff
@@ -0,0 +1,73 @@
...
+
+		return false
+	}
+
+	log := s.log.With().Uint64("slot", uint64(duty.Slot())).Logger()
+	log.Trace().Msg("Checking if we should attest")
+
+	if s.attesterActive.Load() {
+		log.Trace().Msg("We are the active attester")
+		s.enableAttester(ctx)
```
Bez625, line 37: Is it necessary to call `enableAttester` here instead of just `monitorActive`?

services/multiinstance/staticdelay/shouldattest.go
```diff
@@ -0,0 +1,73 @@
...
+		s.disableAttester(ctx)
+
+		return true
+	}
+
+	for _, attestation := range resp.Data {
+		if attestation.AggregationBits.BitAt(validatorCommitteeIndex) {
+			// An attestation is already in the pool; we don't need to act.
+			log.Trace().Msg("Another instance is attesting; not activating attester")
+			s.disableAttester(ctx)
```
Bez625, line 63: Do we need to call `disableAttester` and not just `monitorActive`?

services/multiinstance/staticdelay/shouldattest.go
```diff
@@ -0,0 +1,73 @@
...
+	slot := duty.Slot()
+	committeeIndex := duty.CommitteeIndices()[0]
+	validatorCommitteeIndex := duty.ValidatorCommitteeIndices()[0]
+	resp, err := s.attestationPoolProvider.AttestationPool(ctx, &api.AttestationPoolOpts{
+		Slot:           &slot,
+		CommitteeIndex: &committeeIndex,
+	})
+	if err != nil {
+		log.Warn().Err(err).Msg("Failed to obtain attester duties; activating attester")
+		s.disableAttester(ctx)
```
Bez625, line 54: The error message and return imply we are activating the attester, but we are calling `disableAttester` and I think the intent is to call `enableAttester`? 

Even so I wonder if this is the right thing to do in the event we are not the active vouch and we don't know if the other instance is active. Is it better for us to not attest at all rather than both vouch instances attest and potentially risk slashing?

services/multiinstance/staticdelay/shouldpropose.go
```diff
@@ -0,0 +1,60 @@
+// Copyright © 2024 Attestant Limited.
```
Bez625, line 1:
```suggestion
// Copyright © 2025 Attestant Limited.
```

services/multiinstance/staticdelay/shouldpropose.go
```diff
@@ -0,0 +1,60 @@
...
+
+		return false
+	}
+
+	log := s.log.With().Uint64("slot", uint64(duty.Slot())).Logger()
+	log.Trace().Msg("Checking if we should propose")
+
+	if s.proposerActive.Load() {
+		log.Trace().Msg("We are the active proposer")
+		s.enableProposer(ctx)
```
Bez625, line 38: Do we not just need to call `monitorActive` here?

services/multiinstance/staticdelay/shouldpropose.go
```diff
@@ -0,0 +1,60 @@
...
+	})
+	if err != nil {
+		log.Warn().Err(err).Msg("Failed to obtain beacon block header; activating proposer")
+		s.enableProposer(ctx)
+
+		return true
+	}
+
+	log.Trace().Msg("Beacon block found; not activating proposer")
+	s.disableProposer(ctx)
```
Bez625, line 57: Do we not just need to call `monitorActive` here?

services/multiinstance/staticdelay/shouldattest.go
```diff
@@ -0,0 +1,73 @@
...
+
+	slot := duty.Slot()
+	committeeIndex := duty.CommitteeIndices()[0]
+	validatorCommitteeIndex := duty.ValidatorCommitteeIndices()[0]
+	resp, err := s.attestationPoolProvider.AttestationPool(ctx, &api.AttestationPoolOpts{
+		Slot:           &slot,
+		CommitteeIndex: &committeeIndex,
+	})
+	if err != nil {
+		log.Warn().Err(err).Msg("Failed to obtain attester duties; activating attester")
```
Bez625, line 53: I think the message is slightly inaccurate.

```suggestion
		log.Warn().Err(err).Msg("Failed to obtain attestation pool; activating attester")
```

Bez625 (changes requested):

services/multiinstance/staticdelay/shouldattest.go
mcdee: Yes we should be calling `enableAttester`.

In general we will fail to be active rather than inactive.  Slashing protection is carried out in the signers  What we're trying to avoid with multi-instance is having too many "signing denied" style errors showing up in Vouch, but not at the expense of failing to sign.  So multiple instances signing is a better situation than no instances signing.

services/multiinstance/staticdelay/shouldattest.go
mcdee: It's possible that the proposer could be enabled, so we do want to explicitly call `disableAttester()` even if we think that we're already disabled.

services/multiinstance/always/metrics.go
```diff
@@ -0,0 +1,81 @@
+// Copyright © 2024 Attestant Limited.
```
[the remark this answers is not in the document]
mcdee: This code was written in 2024, so we'll keep the older date in this and similar files.

services/multiinstance/always/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// monitorActive provides metrics for the instance activity status.
+func monitorActive(operation string, active bool) {
+	if activeCounter == nil {
+		return
+	}
+
+	if active {
+		activeCounter.WithLabelValues(operation, "active").Inc()
```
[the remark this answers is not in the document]
mcdee: It's technically accurate, and allows for checking of the operation of the service regardless of the implementation.

services/multiinstance/always/metrics.go
mcdee: True; I've changed the signature to not take the `active` parameter.

services/multiinstance/staticdelay/shouldattest.go
mcdee: It's possible that the proposer could be disabled, so we do want to explicitly call `enableAttester()` even if we think that we're already enabled.

services/multiinstance/staticdelay/shouldpropose.go
mcdee: As per above.

services/multiinstance/staticdelay/shouldpropose.go
mcdee: As per above.

PR #319, opened 2025-04-07
Scan all expected attestations rather than just the first.

mcdee:


mcdee:
This improves the reliability of the multi-instance detection mechanism.

Bez625 (approved):

PR #334, opened 2025-07-17
Fix sync committee preparation by deduplicating signing requests

hoanhan101:
## Description

This PR fixes an issue in sync committee preparation where duplicate signing requests were being sent to multi-signers. When the same account needs to sign the same root multiple times (which can happen during sync committee duties), the previous implementation would send multiple identical signing requests, potentially causing performance issues or errors in the signing infrastructure.

The fix implements deduplication logic that identifies unique (account, root) pairs before sending signing requests, then maps the results back to the original positions to maintain the expected response structure.

## Changes

- Add deduplication logic in `signRootsMulti` function to identify unique (account, root) pairs
- Maintain mapping between original request positions and deduplicated requests
- Map signatures back to original positions after signing
- Add comprehensive tests for the deduplication logic covering various scenarios including no duplicates, exact duplicates, and mixed patterns

### DIFF
--- a/services/signer/standard/helpers.go
+++ b/services/signer/standard/helpers.go
@@ -80,14 +80,46 @@ func (*Service) signRootsMulti(ctx context.Context,
 	}
 
 	if multiSigner, isMultiSigner := accounts[0].(e2wtypes.AccountProtectingMultiSigner); isMultiSigner {
+		// Deduplicate (account, root) pairs to avoid duplicate signing requests
+		type accountRootPair struct {
+			accountKey string
+			rootKey    string
+		}
+
+		uniquePairs := make(map[accountRootPair]int) // map to first occurrence index
+		uniqueAccounts := []e2wtypes.Account{}
+		uniqueData := [][]byte{}
+		originalToUniqueIndex := make([]int, len(accounts)) // maps original index to unique index
+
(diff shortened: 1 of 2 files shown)


Bez625 (changes requested):
Overall looks like a good start!

services/signer/standard/helpers.go
```diff
@@ -165,3 +172,54 @@ func (s *Service) signRootsByAccountType(ctx context.Context, accounts []e2wtype
...
+
+// mapSignaturesToOriginalPositions maps unique signatures back to the original request positions.
+func mapSignaturesToOriginalPositions(
+	sigs []phase0.BLSSignature,
+	uniqueSigs []e2types.Signature,
+	dedup deduplicationResult,
+) {
+	for i := range sigs {
+		uniqueIndex := dedup.originalToUniqueIndex[i]
+		if uniqueIndex < len(uniqueSigs) && uniqueSigs[uniqueIndex] != nil {
```
Bez625, line 221: If any of these conditions are not met we will silently ignore that signature. Is there a reason for this? If not I think we should return an appropriate error.

Bez625 (changes requested):

Bez625 (approved):

PR #391, opened 2026-03-25
Add provider label to block_receipt_delay_seconds histogram

AntiD2ta:
## Summary

- Adds a `provider` label to the `vouch_block_receipt_delay_seconds` histogram, identifying which beacon node delivered the head event
- The address is resolved at event time via a type assertion on the existing events provider to `eth2client.Service`, so it reflects the currently active client even after failovers

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./services/controller/standard/...` — all 29 tests pass (including 2 new)
- [x] `golangci-lint run` — no new issues
- [ ] Manual: run Vouch with Prometheus, query `vouch_block_receipt_delay_seconds_bucket` and confirm `provider` label appears

### DIFF
--- a/services/controller/standard/events.go
+++ b/services/controller/standard/events.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	eth2client "github.com/attestantio/go-eth2-client"
 	"github.com/attestantio/go-eth2-client/api"
 	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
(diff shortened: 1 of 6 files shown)


services/controller/standard/events.go
```diff
@@ -50,7 +51,11 @@ func (s *Service) HandleHeadEvent(ctx context.Context, data *apiv1.HeadEvent) {
 	s.lastBlockRoot = data.Block
 	epoch := s.chainTimeService.SlotToEpoch(data.Slot)
 
-	monitorBlockDelay(uint(uint64(data.Slot)%s.slotsPerEpoch), time.Since(s.chainTimeService.StartOfSlot(data.Slot)))
+	var provider string
+	if svc, ok := s.eventsProvider.(eth2client.Service); ok {
+		provider = svc.Address()
```
Bez625, line 56: I'm pretty convinced this will not work as expected. If you take a look [here](https://github.com/attestantio/go-eth2-client/blob/master/multi/service.go#L139), the multi-client only returns the first active client in the array. 

I think this might require work in go-eth2-client to pass back the provider in the head event data struct. This should tell us the actual provider that sent the event.

Bez625 (changes requested):

PR #392, opened 2026-03-29
Add buildx-compatible multi-arch Docker build for amd64 and arm64

dahu33:
## Summary
This PR updates the Docker build so the image can be built for both `linux/amd64` and `linux/arm64` with `docker buildx`.

This partially resolves #387.

## What changed
- build the Go builder stage on `$BUILDPLATFORM`
- select the appropriate C cross-compiler based on `BUILDARCH`/`TARGETARCH`
- keep the same-arch path on the native compiler
- keep the runtime image on `debian:bookworm-slim` so the PR does not introduce a breaking runtime-base change
- tighten `.dockerignore` so local binaries, VCS metadata, and editor noise are not sent in the Docker build context

## Scope note
This PR only addresses Dockerfile-level multi-arch compatibility.

The arm64/multi-arch behavior comes from the builder-stage and cross-compilation changes. The earlier distroless runtime switch has been reverted because it is an independent breaking change and should not be bundled into this PR.

I could not find the image publish automation in this repository, so this does not update the build-and-push path for the published Docker image. If publishing is handled externally, that external pipeline will still need to build and publish a multi-arch manifest for the arm64 image to appear publicly.

## Verification
- `docker buildx build --platform linux/amd64 --output type=cacheonly .`
- `docker buildx build --platform linux/arm64 --output type=cacheonly .`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 *.dll
 *.so
 *.dylib
+vouch
 
 # Test binary, build with `go test -c`
 *.test
@@ -19,5 +20,16 @@ coverage.html
 
 # Local TODO
 TODO.md
+
+# VCS and editor metadata
+.git
+.gitignore
+.github
+.idea
+.vscode
+.DS_Store
+
+# Logs
+*.log
 
 Dockerfile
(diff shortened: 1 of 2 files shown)


dahu33:
@Bez625 @AntiD2ta any chance someone from your team could review this PR? Thank you!

PR #397, opened 2026-04-09
Use block headers in cache handleHead, full blocks are best-effort now

### LINKED ISSUE #393: Cache service continuously fails to fetch signed beacon blocks by root when block body has been pruned by Lighthouse
### Environment

- **Vouch version**: v1.12.1
- **Beacon node**: Lighthouse v8.1.3-176cce5 (aarch64-linux)
- **Network**: Ethereum Mainnet

### Description

The cache service continuously attempts to fetch signed beacon blocks via `/eth/v2/beacon/blocks/<root>` for block roots that correspond to epoch checkpoint roots (e.g. justified/finalized checkpoint roots). Lighthouse prunes full block bodies for older blocks but retains block headers, causing these requests to return 404 NOT_FOUND on every slot. This results in persistent `warn` and `error` level log entries that never recover.

The interesting observation is that the same block root can be successfully retrieved via `/eth/v1/beacon/headers/<root>`, which only requires the block header. This confirms the blocks exist in Lighthouse's chain, but their bodies have been pruned.

### Steps to Reproduce

1. Run Vouch v1.12.1 against Lighthouse v8.1.3 with default pruning settings
2. Observe the cache service attempting to fetch signed beacon blocks by root
3. The queried roots are epoch checkpoint roots (appearing as `source.root` or `target.root` in attestation data), not the current head block root
4. Lighthouse returns 404 for `/eth/v2/beacon/blocks/<root>` because the block body has been pruned
5. The same root successfully returns data via `/eth/v1/beacon/headers/<root>`

### Observed Logs

Persistent errors appearing on every slot:

```
{"level":"debug","service":"client","impl":"http","address":"http://beacon1:5052",
"endpoint":"/eth/v2/beacon/blocks/0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598",
"status_code":404,"response":{"code":404,"message":"NOT_FOUND: beacon block with root 0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598"},
"message":"GET failed"}

{"level":"warn","strategy":"signedbeaconblock","impl":"first","strategy_id":"83b44c4",
"message":"Failed to obtain signed beacon block before timeout"}

{"level":"error","service":"cache","impl":"standard",
"error":"failed to obtain signed beacon block before timeout",
"message":"Failed to obtain block"}
```

### Verification

The failing block root `0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598` corresponds to slot 14045536, which can be confirmed via the headers endpoint:

```bash
$ curl -s http://beacon1:5052/eth/v1/beacon/headers/0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598
{
  "data": {
    "root": "0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598",
    "canonical": true,
    "header": {
      "message": {
        "slot": "14045536",
        "proposer_index": "491559",
        ...
      }
    }
  }
}
```

But the full block body endpoint returns 404:

```bash
$ curl -s http://beacon1:5052/eth/v2/beacon/blocks/0x5245427e55f3a453c4d77febeb558306acb546b2f1732080bf267442bbc23598
{"code":404,"message":"NOT_FOUND: beacon block with root 0x5245..."}
```

This block root appears as an epoch checkpoint root (e.g., `source.root` in attestation data for epoch 438923), not as the current head block root. The head slot at the time was ~14045658, meaning slot 14045536 is approximately 122 slots (~24 minutes) behind the head — within the range that Lighthouse's default pruning settings would remove the block body.

### Root Cause Analysis

The cache service appears to use `/eth/v2/beacon/blocks/<root>` to resolve block roots to slot numbers for the `BlockRootToSlotCache`. However, when the block root comes from an epoch checkpoint (finalized/justified root), the corresponding block body may already be pruned by Lighthouse, even though the block is recent (within the last few epochs).

A potential fix would be to use `/eth/v1/beacon/headers/<root>` instead of `/eth/v2/beacon/blocks/<root>` when only the slot number is needed for cache population, since block headers are retained by Lighthouse even after block body pruning.

### Impact

- Persistent `warn` and `error` level logs on every slot, making log monitoring noisy
- `BlockRootToSlotCache` fails to update for checkpoint roots
- No direct impact on attestation correctness has been observed, but the continuous errors make it difficult to distinguish real issues from this background noise

### Additional Notes

- Both beacon nodes in the setup (beacon1 and beacon3, both Lighthouse v8.1.3) exhibit the same behaviour
- Changing `beaconblockproposal` strategy from `best` to `first` does **not** resolve the issue, as the cache service operates independently
- Beacon nodes are fully synced (`sync_distance: 0`) at the time the errors occur
- The `beaconblockproposal best` strategy was previously using `beacon2` running Erigon/Caplin v3.3.9, which returned `execution_requests.deposits.amount` as a JSON `number` instead of `string`, causing a separate unmarshal error. Switching to Lighthouse resolved that issue but exposed this pruning-related cache failure.

AntiD2ta:
## Problem

Vouch's cache `handleHead` fetches full signed beacon blocks via `/eth/v2/beacon/blocks/<root>` on every head event. Lighthouse prunes block bodies by default, returning 404 for this endpoint. The `signedbeaconblock/first` strategy silently ignores 404s but never sends a response, causing a timeout and persistent error-level logs on every slot. The execution data extracted from the full block (gas limit, chain head) is only used for builder bid verification and is not critical for core validator duties.

## Solution

Use `data.Slot` from the `HeadEvent` directly for the root-to-slot cache (consistent with `handleBlock`), avoiding an HTTP round-trip entirely. Attempt the full block fetch as best-effort for execution payload data, logging failures at debug level instead of error. The startup path in `service.go:New` uses the `BeaconBlockHeader` endpoint since no event data is available at that point.

Resolves #393.

Dependends on https://github.com/attestantio/vouch/pull/394 which depends on on https://github.com/attestantio/vouch/pull/396 for passing the Trivy scan.

### DIFF
--- a/services/cache/standard/events.go
+++ b/services/cache/standard/events.go
@@ -33,17 +33,7 @@ func (s *Service) handleBlock(_ context.Context, data *apiv1.BlockEvent) {
 func (s *Service) handleHead(ctx context.Context, data *apiv1.HeadEvent) {
 	s.log.Trace().Stringer("root", data.Block).Uint64("slot", uint64(data.Slot)).Msg("Received head event")
 
-	// Use the header endpoint which is always available, even when
-	// block bodies have been pruned by the beacon node.
-	headerResponse, err := s.beaconBlockHeadersProvider.BeaconBlockHeader(ctx, &api.BeaconBlockHeaderOpts{
-		Block: data.Block.String(),
-	})
-	if err != nil {
-		s.log.Warn().Err(err).Msg("Failed to obtain block header")
-		return
-	}
-
-	s.SetBlockRootToSlot(data.Block, headerResponse.Data.Header.Message.Slot)
+	s.SetBlockRootToSlot(data.Block, data.Slot)
 
 	// Best-effort: try to fetch the full block for execution payload data.
 	// This may fail if the beacon node has pruned the block body (e.g. Lighthouse
@@ -53,7 +43,7 @@ func (s *Service) handleHead(ctx context.Context, data *apiv1.HeadEvent) {
 		Block: data.Block.String(),
 	})
 	if err != nil {
-		s.log.Debug().Err(err).Msg("Failed to obtain full block for execution data; block body may be pruned")
+		s.log.Debug().Err(err).Msg("Failed to obtain full block for execution data")
 		return
 	}
 
--- a/services/cache/standard/events.go
+++ b/services/cache/standard/events.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 - 2025 Attestant Limited.
+// Copyright © 2022 - 2026 Attestant Limited.
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -33,16 +33,31 @@ func (s *Service) handleBlock(_ context.Context, data *apiv1.BlockEvent) {
 func (s *Service) handleHead(ctx context.Context, data *apiv1.HeadEvent) {
 	s.log.Trace().Stringer("root", data.Block).Uint64("slot", uint64(data.Slot)).Msg("Received head event")
 
+	// Use the header endpoint which is always available, even when
+	// block bodies have been pruned by the beacon node.
+	headerResponse, err := s.beaconBlockHeadersProvider.BeaconBlockHeader(ctx, &api.BeaconBlockHeaderOpts{
+		Block: data.Block.String(),
+	})
+	if err != nil {
+		s.log.Warn().Err(err).Msg("Failed to obtain block header")
+		return
+	}
+
+	s.SetBlockRootToSlot(data.Block, headerResponse.Data.Header.Message.Slot)
+
+	// Best-effort: try to fetch the full block for execution payload data.
+	// This may fail if the beacon node has pruned the block body (e.g. Lighthouse
+	// default settings).  Execution data is used for builder bid gas limit
+	// verification and is not critical for core validator duties.
 	blockResponse, err := s.signedBeaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
 		Block: data.Block.String(),
 	})
 	if err != nil {
-		s.log.Error().Err(err).Msg("Failed to obtain block")
+		s.log.Debug().Err(err).Msg("Failed to obtain full block for execution data; block body may be pruned")
 		return
 	}
-	block := blockResponse.Data
 
-	s.updateFromBlock(block)
+	s.updateFromBlock(blockResponse.Data)
 }
 
 func (s *Service) updateFromBlock(block *spec.VersionedSignedBeaconBlock) {
--- a/strategies/signedbeaconblock/first/signedbeaconblock.go
+++ b/strategies/signedbeaconblock/first/signedbeaconblock.go
@@ -95,7 +95,7 @@ func (s *Service) SignedBeaconBlock(ctx context.Context,
 	select {
 	case <-ctx.Done():
 		cancel()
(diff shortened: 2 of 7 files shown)


strategies/signedbeaconblock/first/signedbeaconblock.go
```diff
@@ -61,6 +61,7 @@ func (s *Service) SignedBeaconBlock(ctx context.Context,
 			log := log.With().Str("provider", name).Str("block_id", opts.Block).Logger()
 
 			response, err := provider.SignedBeaconBlock(ctx, opts)
+			s.clientMonitor.ClientOperation(name, "signed beacon block", err == nil, time.Since(started))
```
Bez625, line 64: If I'm reading this correctly, we will now increment metrics even if the context is cancelled. Is there a reason we've widened the monitoring for this?

services/cache/standard/events.go
```diff
@@ -33,16 +33,21 @@ func (s *Service) handleBlock(_ context.Context, data *apiv1.BlockEvent) {
...
+	// Best-effort: try to fetch the full block for execution payload data.
+	// This may fail if the beacon node has pruned the block body (e.g. Lighthouse
+	// default settings).  Execution data is used for builder bid gas limit
+	// verification and is not critical for core validator duties.
 	blockResponse, err := s.signedBeaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
 		Block: data.Block.String(),
 	})
 	if err != nil {
-		s.log.Error().Err(err).Msg("Failed to obtain block")
+		s.log.Debug().Err(err).Msg("Failed to obtain full block for execution data")
```
Bez625, line 46: Error -> Debug feels like quite a downgrade. Do we see these errors frequently?

Bez625 (changes requested):

services/cache/standard/events.go
AntiD2ta: If the events provider is a BN that prunes the block body (i.e Lighthouse), then we will witness this log in **every single head event**. We fixed the cache not getting data by using the `header` endpoint, and still try to get the block body in an attempt to fetch execution payload data used for builder bid gas limit verification.

Atm, our cache is always populated with the slot and block header data, and the operators won't see the 404 anymore, but since fetching the block body may fail depending on your BN, I chose Debug so this error (which would be expected) doesn't pollute the logs. Warn is another option, but we might risk seeing a Warn log for every single head event if we use such BNs as the events provider.

strategies/signedbeaconblock/first/signedbeaconblock.go
AntiD2ta: You're reading it correctly. 

The motivation was to align this strategy with every other first strategy ([attestationdata](https://github.com/attestantio/vouch/blob/master/strategies/attestationdata/first/attestationdata.go#L54), beaconblockheader, beaconblockroot, etc.), which all record ClientOperation immediately after the provider call regardless of the error type. Before this change, 404 and 503 responses bypassed monitoring entirely, which made beacon nodes with pruned block bodies (Lighthouse default) invisible in Prometheus, which is what I wanted to fix.

You're right that a side effect is that `context.Canceled` now also gets recorded as a failed operation. I considered that acceptable because:

(a) it matches the behavior of the sibling strategies, and 
(b) ClientOperation takes err == nil as its success flag, so cancellations will show up as failures rather than successes, which is arguably correct since the provider didn't actually return data.

That said, if you'd prefer the cancellation case stay out of the metrics, I'm happy to keep the call where it was and add a separate ClientOperation only inside the 404/503 branches. Let me know which you'd prefer.

Bez625 (approved):

PR #398, opened 2026-04-09
Downgrade timeout and provider failure logs in all first strategies

AntiD2ta:
## Problem

The "first" strategies logged at Warn level for both timeouts and individual provider failures. Since all callers already handle the returned error at their own appropriate log level, the strategy's internal Warn was redundant and created noise — particularly in environments where providers commonly return 404 (Lighthouse block pruning) or 503.

## Solution

Downgrade `log.Warn()` to `log.Debug()` for:
- **Timeout logs** in all 6 remaining first strategies (beaconblockheader, beaconblockroot, beaconblockproposal, attestationdata, aggregateattestation, synccommitteecontribution)
- **Individual provider failure logs** in the 5 older strategies that still used Warn (beaconblockroot, beaconblockproposal, attestationdata, aggregateattestation, synccommitteecontribution)

This makes all 7 first strategies consistent — `signedbeaconblock/first` and `beaconblockheader/first` already used Debug for provider failures.

Depends on #397.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ dev:
   - use block headers as primary fetch in cache handleHead; full blocks are best-effort
   - migrate Dirk and tracing TLS certificate loading to go-certmanager
   - remove dead priorBlocksVotes cache from beaconblockproposal/best strategy
+  - reduce log noise from first strategies on timeouts and provider failures
```
Bez625: Please be a little more explicit and clear with the changelog entries. Noise can be misleading. 

Something like "downgrade warn level log messages to debug for timeouts on first strategies

Bez625 (changes requested):

CHANGELOG.md
AntiD2ta: Done [9fd5207](https://github.com/attestantio/vouch/pull/398/commits/9fd520750e3e0a72528fda3b89017e1ceffa011b)

Bez625 (approved):

PR #401, opened 2026-04-27
Remove dead priorBlocksVotes cache from beaconblockproposal/best strategy

### LINKED ISSUE #400: beaconblockproposal/best HandleHeadEvent logs error on every slot for an unused attestation history cache
### Environment

- **Vouch version**: commit 39ae9714a6f812cb4256c73e77afc673571e3a3d (includes PR #397)
- **Beacon node**: Lighthouse v8.1.3-176cce5 (aarch64-linux)
- **Network**: Ethereum Mainnet
- **Configuration**: `strategies.beaconblockproposal.style: 'best'`

### Summary

Follow-up to #393 / #397. After PR #397 successfully fixed the cache service pruning issue, the `beaconblockproposal/best` strategy exhibits the same underlying problem via a different code path. `HandleHeadEvent` in `strategies/beaconblockproposal/best/events.go` fetches the full signed beacon block on every head event, fails with 404 when Lighthouse has pruned the block body, and emits an `error`-level log on every slot.

The cache this fetch populates (`priorBlocksVotes`) is dead code — it is written to and maintained, but never read. Removing the entire code path eliminates the error with no loss of functionality.

### Observed Logs

Error-level logs appear on every slot, even with `log-level: warn`:

```
Apr 24 11:13:47 vouch[279401]: {"level":"error","strategy":"beaconblockproposal","impl":"best","slot":14181367,"error":"GET failed with status 404: {\"code\":404,\"message\":\"NOT_FOUND: beacon block with root 0x3d4824f1097fe906d66d162a7a66a350e9669c145aa7ca1b1182c8e004d3bf53\",\"stacktraces\":[]}","time":"2026-04-24T11:13:47.925+08:00","message":"Failed to obtain head beacon block"}

Apr 24 11:14:00 vouch[279401]: {"level":"error","strategy":"beaconblockproposal","impl":"best","slot":14181368,"error":"GET failed with status 404: {...}","message":"Failed to obtain head beacon block"}

[... one or two per slot, indefinitely ...]
```

### Root Cause

`HandleHeadEvent` in `strategies/beaconblockproposal/best/events.go` fetches the full signed beacon block via `/eth/v2/beacon/blocks/<root>`:

```go
blockResponse, err := s.signedBeaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
    Block: fmt.Sprintf("%#x", data.Block),
})
if err != nil {
    log.Error().Err(err).Msg("Failed to obtain head beacon block")
    return
}
block := blockResponse.Data

s.updateBlockVotes(ctx, block)
```

This is the same Lighthouse pruning interaction described in #393: Lighthouse retains the block header but prunes the block body for older blocks, so `/eth/v2/beacon/blocks/<root>` returns 404 while `/eth/v1/beacon/headers/<root>` still succeeds.

### `priorBlocksVotes` is dead code

`updateBlockVotes` populates a `priorBlocksVotes` cache containing aggregated attestation bitfields over the last 2 epochs. This cache was introduced in Vouch v1.4.0 (per the release notes: "increase accuracy of beacon block proposal scorer by incorporating attestation history").

A grep across the entire repository confirms the cache is never read:

```
$ grep -rn "priorBlocksVotes" .
./strategies/beaconblockproposal/best/events.go:40:     s.priorBlocksVotesMu.RLock()
./strategies/beaconblockproposal/best/events.go:41:     _, exists := s.priorBlocksVotes[data.Block]
./strategies/beaconblockproposal/best/events.go:42:     s.priorBlocksVotesMu.RUnlock()
./strategies/beaconblockproposal/best/events.go:125:    s.priorBlocksVotesMu.Lock()
./strategies/beaconblockproposal/best/events.go:126:    s.priorBlocksVotes[root] = priorBlockVotes
./strategies/beaconblockproposal/best/events.go:127:    for k, v := range s.priorBlocksVotes {
./strategies/beaconblockproposal/best/events.go:130:                    delete(s.priorBlocksVotes, k)
./strategies/beaconblockproposal/best/events.go:133:    s.priorBlocksVotesMu.Unlock()
./strategies/beaconblockproposal/best/service.go:54:    priorBlocksVotes   map[phase0.Root]*priorBlockVotes
./strategies/beaconblockproposal/best/service.go:55:    priorBlocksVotesMu sync.RWMutex
./strategies/beaconblockproposal/best/service.go:116:           priorBlocksVotes:          make(map[phase0.Root]*priorBlockVotes),
```

All matches are in only two files:

- **`events.go`**: populating and pruning the cache (`HandleHeadEvent` + `updateBlockVotes`)
- **`service.go`**: struct field declaration and map initialization

The current `scoreBeaconBlockProposal` in `strategies/beaconblockproposal/best/score.go` derives its score purely from the `ConsensusValue + ExecutionValue` fields returned by the beacon node in the `VersionedProposal` response:

```go
func (s *Service) scoreBeaconBlockProposal(_ context.Context,
    name string,
    blockProposal *api.VersionedProposal,
) float64 {
    if blockProposal == nil {
        return 0
    }

    score, _ := new(big.Int).Add(blockProposal.ConsensusValue, blockProposal.ExecutionValue).Float64()
    // ...
    return score
}
```

The v1.4.0 scoring logic that consumed `priorBlocksVotes` has been superseded by the newer beacon API contract where the beacon node itself computes and returns the proposal value. The cache-population path was apparently missed when the old scorer was retired.

### Verification

Same as #393 — the failing block root is fetchable via the header endpoint:

```bash
$ curl -s http://beacon-node:5052/eth/v1/beacon/headers/0x3d4824f1097fe906d66d162a7a66a350e9669c145aa7ca1b1182c8e004d3bf53 | jq '.data.header.message.slot'
"14181236"

$ curl -s -o /dev/null -w "%{http_code}\n" http://beacon-node:5052/eth/v2/beacon/blocks/0x3d4824f1097fe906d66d162a7a66a350e9669c145aa7ca1b1182c8e004d3bf53
404
```

### Impact

- Error-level log entries on every slot (1-2 per slot, indefinitely)
- Makes `log-level: warn` unusable for surfacing real issues
- `/eth/v2/beacon/blocks/<root>` is called and its payload parsed on every head event, purely to populate an unused cache — small but nonzero CPU, memory, and network overhead
- No functional impact on block proposal scoring (the cache is never consumed)

### Proposed Fix

Since the cache has no consumers, the cleanest fix is to remove the unused machinery entirely:

- Delete `HandleHeadEvent` and `updateBlockVotes` from `strategies/beaconblockproposal/best/events.go`
- Delete the `priorBlocksVotes` and `priorBlocksVotesMu` fields from `strategies/beaconblockproposal/best/service.go`
- Delete the `priorBlockVotes` struct definition if it becomes unused after the above
- Remove the head event subscription for this strategy if it was registered elsewhere

### Related

- Follow-up to #393 / #397
- Same underlying Lighthouse pruning interaction, different code path

AntiD2ta:
## Summary

- Deletes the `priorBlocksVotes` cache and its head-event writer in `strategies/beaconblockproposal/best`. The cache had **zero readers** — `gopls references` against the field declaration confirmed the only accesses were write-side dedupe and eviction within the writer itself. It was a v1.4.0 fossil left behind when scoring was rewritten to use `VersionedProposal.ConsensusValue + ExecutionValue` directly.
- Stops the eager full-block fetch on every head event. On Lighthouse-style beacon nodes that prune block bodies but retain headers, the fetch returned 404 for blocks within the 2-epoch window, producing one or two `error`-level log entries per slot indefinitely. This made `log-level: warn` unusable.
- Removes the now-unused `signedBeaconBlockProvider` and `eventsProvider` parameters from the strategy and their wiring in `main.go`.
- Drops the now-unused `signedBeaconBlockProvider` parameter from the `selectProposalProvider` → `startProviders` → `startSigningServices` chain in `main.go`. The local `signedBeaconBlockProvider` variable in `start()` is unchanged — it is still passed to `startSharedServices` and `initController`.

Fixes #400. Companion to #397, which fixed the same pattern in `services/cache/standard`.

## Changes

| File | Change |
|------|--------|
| `strategies/beaconblockproposal/best/events.go` | deleted |
| `strategies/beaconblockproposal/best/events_internal_test.go` | deleted |
| `strategies/beaconblockproposal/best/service.go` | drop cache fields, struct, init, head-event subscription, unused imports |
| `strategies/beaconblockproposal/best/parameters.go` | drop `eventsProvider` / `signedBeaconBlockProvider` fields, options, and nil checks |
| `strategies/beaconblockproposal/best/service_test.go` | drop `EventsProviderMissing`, `SignedBeaconBlockProviderMissing`, `ErroringEventsProvider` cases; strip dead options from rest |
| `strategies/beaconblockproposal/best/beaconblockproposal_test.go` | strip dead options |
| `main.go` | remove `WithEventsProvider` / `WithSignedBeaconBlockProvider` from the `best` strategy `New(...)` call; drop unused `signedBeaconBlockProvider` parameter from `selectProposalProvider`, `startProviders`, and `startSigningServices` |
| `CHANGELOG.md` | dev: bullet |

## Out of scope

`services/controller/standard/events.go` has a similar event-driven full-block fetch in `VerifySyncCommitteeMessages`, but it is **not** the same bug: it logs at Debug, has a real consumer (`ParentRoot()` and `SyncAggregate()` for sync-committee inclusion verification), and fetches the current head (rarely pruned).

## Test plan

- [x] Devnet: run vouch against a Lighthouse beacon node with `log-level: warn` and the `best` beaconblockproposal strategy. Confirm no more `Failed to obtain head beacon block` errors per slot, and that block proposals still score and are selected. User local block building
- [ ] Testnet (Hoodi): similar setup as above


strategies/beaconblockproposal/best/service.go
```diff
@@ -32,15 +29,14 @@ import (
...
+	clientMonitor          metrics.ClientMonitor
+	processConcurrency     int64
+	chainTime              chaintime.Service
+	proposalProviders      map[string]eth2client.ProposalProvider
+	timeout                time.Duration
+	blockRootToSlotCache   cache.BlockRootToSlotProvider
+	executionPayloadFactor float64
 
 	// Spec values for scoring proposals.
 	slotsPerEpoch      uint64
```
xav: Looks like `slotsPerEpoch` is not used anywhere now that `events.go` is deleted. In that case, maybe we should remove this field too?

strategies/beaconblockproposal/best/service.go
```diff
@@ -98,36 +83,24 @@ func New(ctx context.Context, params ...Parameter) (*Service, error) {
...
 	s := &Service{
-		log:                       log,
-		processConcurrency:        parameters.processConcurrency,
-		chainTime:                 parameters.chainTime,
-		proposalProviders:         parameters.proposalProviders,
-		signedBeaconBlockProvider: parameters.signedBeaconBlockProvider,
-		timeout:                   parameters.timeout,
-		blockRootToSlotCache:      parameters.blockRootToSlotCache,
-		clientMonitor:             parameters.clientMonitor,
-		slotsPerEpoch:             scoringData.slotsPerEpoch,
```
xav, line 109: Same here as for `Service`, can we remove this field?

AntiD2ta: Done [6db7a1a](https://github.com/attestantio/vouch/pull/401/commits/6db7a1a7e0070a4aed54772af3bb36bc0870f774)

strategies/beaconblockproposal/best/service.go
AntiD2ta: Good catch, done [6db7a1a](https://github.com/attestantio/vouch/pull/401/commits/6db7a1a7e0070a4aed54772af3bb36bc0870f774)

Bez625 (approved):

PR #416, opened 2026-07-23
Add lock around sync messenger map access

AntiD2ta:
Backport of #406 to the `v1.13.x` release branch. Adds locking around the sync committee messenger's in-memory `slotDataRecords` map to fix a data race.

Previously:

- `GetDataUsedForSlot` read `slotDataRecords` with **no lock at all** while `UpdateSyncCommitteeDataRecord` could write concurrently.
- `RemoveHistoricDataUsedForSlotVerification` performed its `len()` check and map iteration outside the lock (only the inner `delete` loop was guarded).

A concurrent map read/write in Go triggers a fatal `concurrent map read and map write` runtime panic, which would crash the whole Vouch process — a liveness/DoS risk.


Bez625 (approved):

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.43.0
 	go.uber.org/atomic v1.11.0
 	golang.org/x/sync v0.22.0
-	google.golang.org/grpc v1.80.0
+	google.golang.org/grpc v1.82.1
```
Bez625, line 40: Why have you bumped this?

Bez625 (approved):

PR #420, opened 2026-08-11
Add Gloas self-build proposal support

AntiD2ta:
## Summary

- Add the Gloas-only self-build proposal path: request an ePBS proposal with its payload, validate it, sign the beacon block and matching execution-payload envelope, then publish the block followed by its envelope.
- Keep the pre-Gloas relay auction and blinded/unblinding flow unchanged.
- Extend the first and best proposal strategies, signing/submission layers, cache handling, and telemetry for ePBS data.
- Add an opt-in custom beacon-node preset setting for networks that need dynamic SSZ support.

## Reviewer guide

1. Start in `services/beaconblockproposer/standard/propose.go`. The fork gate chooses the self-build path. Its key invariants are an included payload, the expected proposer and slot, a non-zero fee recipient, a correct block root, and completion of both signatures before anything is submitted.
2. Follow `services/signer` and `services/submitter`. The block is still submitted before its envelope, but the envelope is fully prepared first; multinode submission lets all configured nodes receive it within the existing timeout.
3. Review `strategies/beaconblockproposal/{first,best}`. They request ePBS proposals, reject malformed or payload-excluded responses when payload inclusion was requested, and preserve provider-local graffiti handling.
4. Then inspect `clients.go`, `docs/configuration.md`, and the go-eth2-client bump for custom-preset decoding and preset-aware body-root handling.
5. Finally, review cache and telemetry changes; Gloas signed blocks do not update execution-chain state because an ePBS bid has no execution block number.

## Fork schedule lookup and performance

The proposal fork gate still uses `chainTime.HardForkEpoch(...)`, but chain time now publishes an immutable snapshot of every `*_FORK_EPOCH` value returned by the beacon node's spec. Known forks use one atomic snapshot read and a map lookup: no mutex, allocation, or beacon-node request is added to `Prepare()` or `proposeBlock()`.

An unknown fork name triggers a generation-coalesced refresh, allowing future hard forks to be discovered without another fork-specific cache field. Refreshes preserve last-known-good entries, reject changes to already activated forks, and publish a new snapshot only when the effective schedule changes.

Suggested review order:

1. `services/chaintime/standard/service.go` for snapshot construction, lookup, and refresh behavior.
2. `service_test.go` for public behavior and failure handling.
3. `service_internal_test.go` for immutable publication and concurrent refresh guarantees.
4. `service_benchmark_test.go` for the lookup benchmark.

Benchmark command:

```console
gosilent test --verbose -run '^$' -bench '^BenchmarkHardForkEpoch$' -benchmem -count=10 ./services/chaintime/standard
```

Apple M4 median across 10 samples:

| `HardForkEpoch` | Previous implementation | Cached snapshot |
| --- | ---: | ---: |
| Time | 12.31 ns/op | 5.24 ns/op |
| Memory | 8 B/op | 0 B/op |
| Allocations | 1 alloc/op | 0 allocs/op |

That is approximately **2.35x faster**, or **57% lower lookup latency**, while eliminating the allocation. The benchmark uses an in-memory provider to isolate lookup overhead; in production, known-fork lookups also avoid the proposal-time beacon-node request made by the previous implementation.

## Risk and validation

- The Gloas path deliberately avoids relay building and uses a zero builder boost factor; operators are warned when Gloas ignores their relay settings.
- Failure validation happens before signing or submission where possible. Missing required signing/submission configuration fails at startup.
- The focused regression tests cover fork selection, proposal validation, signing/submission order, strategy behavior, custom-preset decoding, cache behavior, and telemetry.
- Current GitHub test, lint, and Trivy checks pass. DeepSource remains failing and should be reviewed separately.

### Kurtosis devnet

A local `glam-d7` devnet remained healthy and Vouch received real Gloas proposer duties. During the initial testing iterations, some beacon-node responses lacked an ePBS proposal; others contained an envelope whose beacon-block root did not match the block. Vouch rejected those responses before signing or submitting, exercising the live failure-safe path. Most of the issues were related to differences between mainnet and minimal preset, and Vouch and its dependencies being tightly coupled to mainnet preset. The devnet was running on minimal preset.

A successful end-to-end ePBS publication was later demonstrated after several fixes among Vouch and go-eth2-client.

**Update on 14/09/26**: Devnet 8 has been launched and this PR has not been tested end-to-end with a local devnet based on Devnet 8. Any new local devnet/enclaves will match as much as possible Devnet 8 settings, including a mainnet preset.


services/signer/standard/signexecutionpayloadenvelope.go
```diff
@@ -0,0 +1,61 @@
...
+	defer span.End()
+
+	if envelope == nil {
+		return phase0.BLSSignature{}, errors.New("no execution payload envelope supplied")
+	}
+	if s.beaconBuilderDomainType == nil {
+		return phase0.BLSSignature{}, errors.New("DOMAIN_BEACON_BUILDER unavailable in beacon node spec; cannot sign execution payload envelope")
+	}
+
+	root, err := envelope.HashTreeRoot()
```
Bez625, line 45: The envelope is signed over `envelope.HashTreeRoot()` (the generated static hasher), but in `propose.go` `epbsProposalEnvelope` deliberately avoids that hasher for the block, using `proposal.Root()`/`BodyRoot()` with the comment that the generated hasher "inlines mainnet preset sizes, so it is wrong on any other preset." That care isn't mirrored here. The envelope embeds `ExecutionPayload` and `BlobKZGCommitments` (whose SSZ list depth depends on `MAX_BLOB_COMMITMENTS_PER_BLOCK`), so on a non-mainnet preset this root — and therefore the signature — could be computed against the wrong root and be rejected by the beacon node. Note `custom-spec-support` only changes the *client's* encode/decode, not this local `HashTreeRoot()`.

Could you confirm the envelope root is genuinely preset-independent, or route it through the same transport/dynssz-derived root used for the block? A test that makes the envelope root preset-sensitive (mirroring `TestProposeGloasSignsRetainedBodyRoot`) would close the gap.

services/beaconblockproposer/standard/propose.go
```diff
@@ -177,39 +184,215 @@ func (s *Service) proposeBlock(ctx context.Context,
...
+			Gloas: &gloas.SignedExecutionPayloadEnvelope{
+				Message:   envelope,
+				Signature: signature,
+			},
+		},
+		KZGProofs: kzgProofs,
+		Blobs:     blobs,
+	}
+	var envelopeSubmissionErr error
+	for attempts := 3; attempts > 0; attempts-- {
```
Bez625, line 328: I'm not convinced by this retry loop. I'm happy to revisit it later, but it feels like we're kind of guessing and the guesses are all hard-coded. If we do need to have a retry those params should be configurable. 

Maybe to ask a different question - is there a point in the slot where we know the payload is ready?

services/beaconblockproposer/standard/propose.go
```diff
@@ -177,39 +184,215 @@ func (s *Service) proposeBlock(ctx context.Context,
...
+	// it.  The payload's own fee recipient collects the slot's priority fees and MEV, so
+	// it is the one that has to be checked before this envelope is signed.
+	if envelope.Payload.FeeRecipient.IsZero() {
+		return nil, phase0.Root{}, errors.New("ePBS execution payload envelope has 0 fee recipient")
+	}
+	if proposal.GloasContents == nil || proposal.GloasContents.Block == nil || proposal.GloasContents.Block.Body == nil || proposal.GloasContents.Block.Body.SignedExecutionPayloadBid == nil || proposal.GloasContents.Block.Body.SignedExecutionPayloadBid.Message == nil {
+		return nil, phase0.Root{}, errors.New("ePBS proposal has no execution payload bid")
+	}
+	if proposal.GloasContents.Block.Body.SignedExecutionPayloadBid.Message.BuilderIndex != gloas.BuilderIndex(math.MaxUint64) {
+		return nil, phase0.Root{}, errors.New("ePBS execution payload bid is not self-built")
```
Bez625, line 371: I'm assuming this is a transient change for self-built only.

services/chaintime/standard/service.go
```diff
@@ -135,30 +158,149 @@ func (s *Service) FirstSlotOfEpoch(epoch phase0.Epoch) phase0.Slot {
...
+	// A concurrent refresh may have published the requested fork before we captured its generation.
+	forkSchedule = s.forkSchedule.Load()
+	if forkEpoch, exists := forkSchedule.epochs[hardForkName]; exists {
+		return forkEpoch
+	}
+	if _, exists := forkSchedule.malformed[hardForkName]; exists {
+		s.log.Error().Err(fmt.Errorf("%s is not a uint64", hardForkName)).Msg("Failed to obtain hard fork")
+		return 0xffffffffffffffff
+	}
+	if err := s.refreshForkSchedule(ctx, generation); err != nil {
```
Bez625, line 180: For a network where `GLOAS_FORK_EPOCH` is not yet in the spec (e.g. current mainnet), every proposal duty misses the snapshot and reaches this `refreshForkSchedule`, issuing a `Spec()` request on the proposal path. That's no worse than the previous implementation (which called `Spec()` every time) and only fires on proposal duties, but the PR's "no beacon-node request added" benefit holds only for *known* forks. Can you confirm the client caches `Spec()` so this isn't a real network round-trip each duty? 

Separately: the generation/leader-election/merge machinery is a lot of concurrency surface for a schedule that changes ~never — it's benchmarked and tested, so not blocking, just noting the complexity trade-off.

Overall - I see the attempt here to make fork look-ups more generic, but I do like the simplicity of setting the fork epoch on the Service at start-up rather than looking up (from an admittedly memoized cache). WDYT?

main.go
```diff
@@ -814,8 +815,10 @@ func startSigningServices(ctx context.Context,
 		standardbeaconblockproposer.WithGraffitiProvider(graffitiProvider),
 		standardbeaconblockproposer.WithMonitor(monitor),
 		standardbeaconblockproposer.WithProposalSubmitter(submitterStrategy.(submitter.ProposalSubmitter)),
+		standardbeaconblockproposer.WithExecutionPayloadEnvelopeSubmitter(submitterStrategy.(submitter.ExecutionPayloadEnvelopeSubmitter)),
```
Bez625: These (and the signer assertion just below) are hard type assertions that panic at startup if the strategy/signer doesn't implement the interface. Consistent with the surrounding code, so acceptable, but a comma-ok form with a `log.Fatal` would give a clearer failure message. The runtime `nil` guards in `proposeEPBSBlock` are a good backstop regardless.

strategies/beaconblockproposal/first/service.go
```diff
@@ -33,10 +35,122 @@ import (
...
+) (
+	*api.Response[*api.VersionedEPBSProposal],
+	error,
+) {
+	ctx, span := otel.Tracer("attestantio.vouch.strategies.beaconblockproposal.first").Start(ctx, "EPBSProposal", trace.WithAttributes(
+		attribute.Int64("slot", util.SlotToInt64(opts.Slot)),
+	))
+	defer span.End()
+
+	ctx, cancel := context.WithTimeout(ctx, s.timeout)
```
Bez625, line 54: Nit: `cancel()` is reached via the select branches rather than `defer cancel()`. It's correct (the loop always terminates at accept or timeout, and every path calls `cancel()`), but a `defer cancel()` right after this line would be more obviously leak-free.

strategies/beaconblockproposal/best/beaconblockproposal.go
```diff
@@ -70,7 +319,8 @@ func (s *Service) Proposal(ctx context.Context,
 	errCh := make(chan *beaconBlockError, requests)
 	// Kick off the requests.
 	for name, provider := range s.proposalProviders {
-		providerGraffiti := opts.Graffiti[:]
+		providerOpts := *opts
```
Bez625, line 322: Nice — copying `opts` per provider here (instead of the previous shared-`opts` mutation) also closes a pre-existing bug where a `{{CLIENT}}` graffiti substitution for one provider could leak into another provider's request.

Bez625 (changes requested):

strategies/beaconblockproposal/first/service.go
AntiD2ta: Agreed. Every current branch cancels the context, but defer `cancel()` makes that invariant explicit and protects future return paths. I will change it.

services/chaintime/standard/service.go
AntiD2ta: Confirmed: the HTTP client caches `Spec()`, so a missing fork does not normally cause a network round trip on every proposal duty. The cache is cleared every five minutes, so an occasional refetch remains possible. I propose resolving `GLOAS_FORK_EPOCH` once when the proposer service starts and storing it on that service. 

I agree that the generic refresh machinery is more complexity than this PR needs. Avoiding complexity is appealing to me after bearing in mind all the remaining changes in Vouch that are proposed in this PR stack. I was seduced by avoiding restarting Vouch to change the schedule with an elegant solution, but this is silly since we most likely have to upgrade Vouch in such a case, and restart it anyways (e.g due to hardfork).

services/beaconblockproposer/standard/propose.go
AntiD2ta: Correct. This PR only supports the self-built Gloas path. `UINT64_MAX is BUILDER_INDEX_SELF_BUILD`, and this guard prevents us from signing and publishing an external-builder response if the beacon node ignores the requested zero boost factor.

 Vouch PR #440 changes this conditional in commit [18fda92b](https://github.com/attestantio/vouch/commit/18fda92b556f2f4c169c3e3c828f9e275931b758#diff-a6fd48ef4c80636527dc13ce6946d0a5c9df2c657a806a69a8c257615e309b60L416)

services/beaconblockproposer/standard/propose.go
AntiD2ta: The payload is already ready, so that does not justify the retry. However, there may still be a race between block acceptance and the envelope endpoint recognising its root, particularly with independent multinode fan-out. The current submitter also collapses per-node errors, so we cannot safely distinguish that transient case yet. I propose retaining the bounded retry in this PR and following up with per-node error preservation and a retry limited to unknown-block responses.

services/signer/standard/signexecutionpayloadenvelope.go
AntiD2ta: I added such test on go-eth2-client side: https://github.com/attestantio/go-eth2-client/pull/329

services/beaconblockproposer/standard/propose.go
AntiD2ta: Addressed in [4d22921](https://github.com/attestantio/vouch/pull/420/commits/4d22921b8edf836acc0713dea3f1726072550b0d)

strategies/beaconblockproposal/first/service.go
AntiD2ta: Done in [62aecdd](https://github.com/attestantio/vouch/pull/420/commits/62aecdd31478310a96b5d25752fdf42eedbaa860)

services/chaintime/standard/service.go
AntiD2ta: Addressed in [468495f](https://github.com/attestantio/vouch/pull/420/commits/468495f16aff5c94f36a1197acd4bc47c593b4e5)

main.go
```diff
@@ -1451,7 +1462,21 @@ func selectProposalProvider(ctx context.Context,
...
+		if !isProvider {
+			return nil, errors.New("beacon block proposal client does not support ePBS proposals")
+		}
+		proposalProvider, err = firstbeaconblockproposalstrategy.New(ctx,
+			firstbeaconblockproposalstrategy.WithClientMonitor(monitor.(metrics.ClientMonitor)),
+			firstbeaconblockproposalstrategy.WithLogLevel(util.LogLevel("strategies.beaconblockproposal.first")),
+			firstbeaconblockproposalstrategy.WithProposalProviders(map[string]eth2client.MultiForkProposalProvider{
+				"simple": provider,
+			}),
+			firstbeaconblockproposalstrategy.WithTimeout(util.Timeout("strategies.beaconblockproposal.first")),
```
Bez625, line 1475: The default style now builds a `first` strategy around the multiclient, which I think can be dropped entirely rather than deduplicated.

`selectProposalProvider` returns `eth2client.MultiForkProposalProvider`, and `multi.Service` already satisfies it: it has `ProposalProvider` (asserted in `multi/interfaces.go`) and `EPBSProposal` (`multi/epbsproposal.go`). So the assertion just above succeeds, `proposalProvider = provider` is all this case needs, and the wrapper only adds a goroutine, a channel and a second timeout around a single provider.

That said, I don't think we can fold `simple` into `first` either — they aren't interchangeable. `simple` is one multiclient failing over across the configured addresses; `first` is N separate clients raced in parallel. Wrapping also means a simple-style operator silently inherits `strategies.beaconblockproposal.first.timeout` and log level (`util.Timeout` walks up the key tree), which they never configured.

Unwrapping would also avoid the error-swallowing noted in the comment on `strategies/beaconblockproposal/first/service.go`.

main.go
```diff
@@ -1451,7 +1462,21 @@ func selectProposalProvider(ctx context.Context,
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to fetch clients for simple beacon block proposal strategy")
 		}
-		proposalProvider = beaconBlockProposalClient.(eth2client.ProposalProvider)
+		provider, isProvider := beaconBlockProposalClient.(eth2client.MultiForkProposalProvider)
+		if !isProvider {
+			return nil, errors.New("beacon block proposal client does not support ePBS proposals")
+		}
+		proposalProvider, err = firstbeaconblockproposalstrategy.New(ctx,
```
Bez625, line 1469: **Low (pre-existing, not introduced by this PR) — `{{CLIENT}}` graffiti substitution does not work on this style.**

`obtainGraffiti` (`services/beaconblockproposer/standard/propose.go:127`) substitutes `{{CLIENT}}` only if `s.proposalProvider` implements `consensusclient.NodeClientProvider`. The `first` strategy does not implement it and, unlike `best` (`strategies/beaconblockproposal/best/beaconblockproposal.go:214` and `:323`), does no `{{CLIENT}}` handling of its own — so a configured `vouch/{{CLIENT}}` is published verbatim, braces and all.

To be clear this is not a regression from the new wrapper: the old simple path assigned `fetchMultiClient`'s result, a `go-eth2-client/multi.Service`, which also has no `NodeClient` method (there is no `multi/nodeclient.go`, and `NodeClientProvider` is absent from the assertion block in `multi/interfaces.go`). Substitution was already dead on both `simple` and `first`.

Still worth fixing while this code is being touched — either forward `NodeClientProvider` through the `first` strategy, or give it the same `{{CLIENT}}` handling `best` has.

services/signer/standard/signexecutionpayloadenvelope.go
Bez625: **Medium — envelope signing root uses the mainnet-baked generated hasher.**

This PR adds `eth2client.custom-spec-support` precisely for non-mainnet presets, and go-eth2-client's `VersionedEPBSProposal.BodyRoot()`/`Root()` carry explicit comments that the generated hashers "inline mainnet preset sizes, so recomputing here would silently sign the wrong root on any other preset".

`envelope.HashTreeRoot()` here is that same generated path, and `ExecutionPayloadEnvelope` contains preset-sized lists (`MAX_WITHDRAWALS_PER_PAYLOAD`, `MAX_BLOB_COMMITMENTS_PER_BLOCK`, …). On a minimal or custom preset Vouch would sign a wrong root and every envelope would be rejected — with the block already published, i.e. a lost payload reveal.

This should take the root from the same preset-aware source the block path uses rather than recomputing it here.

strategies/beaconblockproposal/best/beaconblockproposal.go
```diff
@@ -41,6 +41,254 @@ type beaconBlockError struct {
...
+			return errors.New("beacon node returned malformed ePBS proposal")
+		}
+		block = proposal.GloasContents.Block
+	}
+
+	if block == nil || block.Body == nil || block.Body.SignedExecutionPayloadBid == nil || block.Body.SignedExecutionPayloadBid.Message == nil {
+		return errors.New("beacon node returned malformed ePBS proposal")
+	}
+
+	if block.Body.SignedExecutionPayloadBid.Message.FeeRecipient.IsZero() {
```
Bez625: **Medium — rejecting a proposal on a zero *bid* fee recipient discards valid self-built proposals.**

`propose.go:412-418` itself documents that a self-built bid pays nothing and that the *payload's* fee recipient is the meaningful one. The Gloas spec only consumes `bid.fee_recipient` when `amount > 0` (`process_execution_payload_bid`), and self-builds require `value == 0` — so a beacon node is free to leave this field zero.

Since `proposeEPBSBlock` always requests a self-build (`BuilderBoostFactor = 0`), a node that zeroes the unused field makes every provider's response discarded and the slot missed, with `no ePBS proposals received`.

Suggest checking the payload's fee recipient, or only enforcing this when `bid.Message.Value > 0`. Same check exists at `strategies/beaconblockproposal/first/service.go:142`.

Worth confirming against a real beacon node before merge — if any client zeroes this on self-builds, every ePBS slot is missed.

strategies/beaconblockproposal/first/service.go
```diff
@@ -33,10 +34,121 @@ import (
...
+				return false
+			}
+			block = proposal.GloasContents.Block
+		}
+		if block == nil || block.Body == nil || block.Body.SignedExecutionPayloadBid == nil || block.Body.SignedExecutionPayloadBid.Message == nil {
+			s.log.Warn().Msg("Discarding malformed ePBS proposal")
+
+			return false
+		}
+		if block.Body.SignedExecutionPayloadBid.Message.FeeRecipient.IsZero() {
```
Bez625: **Medium — same zero-fee-recipient issue as `strategies/beaconblockproposal/best/beaconblockproposal.go:285`.**

A self-built bid (`value == 0`, which is what `proposeEPBSBlock` always requests) may legitimately carry a zero `bid.fee_recipient` — the Gloas spec only reads that field when `amount > 0`. Discarding on it can drop every valid proposal and miss the slot.

services/beaconblockproposer/standard/propose.go
```diff
@@ -177,39 +186,265 @@ func (s *Service) proposeBlock(ctx context.Context,
...
+	}
+	if envelopeSubmissionErr != nil {
+		return errors.Wrap(envelopeSubmissionErr, "failed to submit execution payload envelope after block publication")
+	}
+	monitorBeaconBlockProposalSource("local")
+
 	return nil
 }
 
+func isUnknownBeaconBlockError(err error) bool {
```
Bez625, line 356: **Medium — this requires *all* multinode errors to be retryable, so mixed failures skip the retry.**

For a `*submitter.SubmissionError` the loop returns `false` as soon as any single node error is not an unknown-block error. `multinode.SubmitExecutionPayloadEnvelope` only returns that aggregate when every node failed, with one wrapped error per node.

With two beacon nodes where node A replies "unknown block" (propagation lag — exactly the case the retry above exists for) and node B fails with a connection error or 500, we break after one attempt and never publish the envelope, even though a 250ms retry against node A would have succeeded.

The timeout path has the same problem: the multinode submitter appends a synthetic `"no successful submissions before timeout"` error into the same set, and that is never an unknown-block error.

`if any(...)` matches the intent better than the current `if all(...)`.

services/submitter/errors.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// NewSubmissionError creates an aggregate of individual submission errors.
+func NewSubmissionError(errs ...error) *SubmissionError {
+	return &SubmissionError{errors: errs}
+}
+
+// Error returns the combined submission error messages.
+func (e *SubmissionError) Error() string {
+	return errors.Join(e.errors...).Error()
```
Bez625, line 30: **Nit — nil dereference when built from an empty or all-nil error set.**

`errors.Join` returns `nil` when every argument is nil (or there are none), so `NewSubmissionError().Error()` panics.

Current call sites in `submitexecutionpayloadenvelope.go` always append at least one non-nil error, but `NewSubmissionError` is exported package API with no guard. A nil check here (or in the constructor) would make it safe.

strategies/beaconblockproposal/first/service.go
```diff
@@ -33,10 +34,121 @@ import (
...
+
+	ctx, cancel := context.WithTimeout(ctx, s.timeout)
+	defer cancel()
+
+	proposalCh := make(chan *api.VersionedEPBSProposal, len(s.proposalProviders))
+	for name, provider := range s.proposalProviders {
+		go s.fetchEPBSProposal(ctx, name, provider, opts, proposalCh)
+	}
+
+	for {
```
Bez625, line 61: **Low — this blocks for the full timeout and loses the node's error when every provider fails.**

`fetchEPBSProposal` returns silently on error, and `acceptableEPBSProposal` rejections `continue`, so this loop can only exit via `ctx.Done()` once nothing usable arrives.

With a single provider — which is now the default, since the `simple` case in `main.go` wraps one multiclient in this strategy — a beacon node that errors instantly makes Vouch burn the whole timeout and then report `failed to obtain ePBS beacon block proposal before timeout` instead of the actual beacon-node error. The default configuration previously surfaced that error directly, so this is a diagnostics regression for exactly the case where it matters, a missed proposal.

Tracking the number of finished providers would let it fail fast with the real cause. The pre-existing `Proposal` method below has the same shape, so the two are worth fixing together.

main.go
AntiD2ta: Agreed on dropping the wrapper, with one adjustment. The wrapper is also carrying ePBS candidate validation that the standard proposer does not fully repeat, including nil-data and zero bid fee-recipient checks.

I plan to move those checks into shared validation called by `first` per candidate and by the standard proposer before it accesses the selected proposal. The default/simple case can then assign the multiclient directly, preserving its failover and errors without inheriting `first`'s timeout or log configuration. `simple` and `first` will remain separate.

main.go
AntiD2ta: Agreed. This is pre-existing, but `first` should handle substitution per provider rather than expose one `NodeClient` for the whole race. I plan to mirror `best`: copy the options for each provider, substitute that provider's client name, and then issue the request. I will cover both proposal endpoints.

The `simple` multiclient case remains separate because `multi.Service` cannot currently identify the eventual proposal client before failover. I do not want to add forwarding that can report a different client from the one that builds the block.

services/beaconblockproposer/standard/propose.go
AntiD2ta: Good catch. The aggregate check currently requires all node failures to be retryable, but one unknown-block response is enough to justify retrying that node after propagation.
 
I plan to change the aggregate case to `any` semantics and add coverage for both unknown-block plus permanent failure and unknown-block plus the multinode timeout marker. Non-aggregate classification will remain unchanged.

strategies/beaconblockproposal/first/service.go
AntiD2ta: Agreed. Each provider currently returns silently on failure or rejection, so the receiver cannot distinguish "all providers finished" from "providers are still pending."

I plan to have every provider send one result, count completions, and return preserved provider errors once all sources finish without an acceptable proposal. The same result path should be used by Proposal and EPBSProposal. A genuinely pending provider will still be bounded by the strategy timeout.

main.go
AntiD2ta: Addressed in https://github.com/attestantio/vouch/commit/382cb50b682b296ea1cb6be613eb3bd911043f44

main.go
AntiD2ta: Addressed in https://github.com/attestantio/vouch/commit/767406a0281d7f64c030686df0732445f1efae81

services/beaconblockproposer/standard/propose.go
AntiD2ta: Addressed in https://github.com/attestantio/vouch/commit/24f378931d5a07927ac950735070fe6d61ec5746

services/submitter/errors.go
AntiD2ta: Addressed in https://github.com/attestantio/vouch/commit/df237866385bd350fc3fe7f108e421bb2a4ac49f

strategies/beaconblockproposal/first/service.go
AntiD2ta: Addressed in https://github.com/attestantio/vouch/commit/29159ad02a5b2e48925b45606754525f86c57e3d

Bez625 (approved):

PR #429, opened 2026-08-18
Add payload timeliness committee duty flow

AntiD2ta:
## Summary

- schedule payload timeliness committee duties through the Gloas payload timing window
- batch generic signatures for validators sharing a slot and submit versioned payload attestations
- add immediate and multinode submission support, reorg refreshes, and payload-slot validation

## Stacked on #428

This branches off `gloas-attestation-deadlines` and targets it, not `gloas`. #428 supplies the
`gloasForkEpoch` field, the spec-derived deadline machinery, and the `dueBPS` helper that the payload
deadlines here are built on, so it should be reviewed and merged first. The diff shown against that
base is this change alone.

## The duty flow

A payload timeliness committee duty is a vote on whether the slot's execution payload was revealed on
time. `PTCDuties` is fetched for the epoch, duties are grouped by slot so that every validator
attesting in the same slot shares one job, and each slot's job runs inside the window the fork
defines:

```go
jobTime := s.chainTimeService.StartOfSlot(duty.Slot()).Add(s.payloadAttestationDelay).Add(payloadAttestationGrace)
deadline := s.chainTimeService.StartOfSlot(duty.Slot() + 1)
```

The offset is derived from the chain specification by `obtainPayloadAttestationTiming`, in the same
basis points of `SLOT_DURATION_MS` that #428 established. On a 12-second slot it is the
specification's `PAYLOAD_ATTESTATION_DUE_BPS: 7500`, so the vote is cast just after 9s and its
context runs to the end of the slot. The duty is new in Gloas, so it always follows the Gloas slot
duration and has no pre-fork form. Why the vote is cast there, rather than when the payload becomes
due, is recorded in the decision section below.

Signing goes through `SignPayloadAttestationData`, which takes the accounts as a slice and returns
one signature per account, so a slot with many attesting validators costs one signing round trip
rather than one per validator.

Submission is versioned and covers all three submitters — `immediate`, `multinode` and `null` — so
the payload vote follows whatever submission strategy is already configured rather than introducing
a separate path.

## Decision: when the payload timeliness vote is cast

Recorded here because the choice is not evident from the constants, and an earlier revision of this
branch made the opposite one.

### Context

Gloas defines two payload deadlines, and they answer different questions.

| Constant | Mainnet value | Role |
| --- | --- | --- |
| `PAYLOAD_DUE_BPS` | 5000 bps — 6s | the *subject* of the vote |
| `PAYLOAD_ATTESTATION_DUE_BPS` | 7500 bps — 9s | the *broadcast deadline* |

`specs/gloas/validator.md` is explicit that the second is a window bound and not a target instant: a
validator "should create and broadcast the `payload_attestation_message` ... within the first
`get_payload_attestation_due_ms()` milliseconds of the slot". Any moment between seeing the block and
that bound is conformant.

The first is not a scheduling instruction at all — it is baked into the vote's content.
`data.payload_present` is `True` only if a `SignedExecutionPayloadEnvelope` "was seen before
`get_payload_due_ms()` milliseconds into the slot". That is a question about the past, its answer is
frozen once the payload due time passes, and the observation timestamp that answers it belongs to the
beacon node rather than to us.

Two further fields have no time cutoff whatsoever: `data.blob_data_available` is
`is_data_available(root)` evaluated when the data is produced, and the duty is skipped altogether if
no beacon block for the slot has been seen. Both only become more accurate later in the slot.

### Why the payload due time was the wrong place

The earlier revision scheduled the job at `PAYLOAD_DUE_BPS` and used `PAYLOAD_ATTESTATION_DUE_BPS`
only as the job's context deadline, reasoning that the answer is frozen at the payload due time so
there is nothing to gain by waiting. That holds for `payload_present` in isolation and fails for the
request as a whole.

- **A beacon node need not answer yet.** Prysm's producer returns `Unavailable` — a 503 — when asked
  before the attestation deadline unless the answer is already final, on the stated grounds that
  "before the deadline only the final result is safe to return ... otherwise both flags may still
  flip". Asking at the payload due time therefore returns nothing in precisely the marginal cases the
  vote exists to judge, and our fetch is a one-shot: the duty ends with an error log and no vote.
- **Data availability is decided later.** A payload revealed near the due time will often still have
  columns propagating, so a vote cast then reports `blob_data_available: false` for a payload that was
  in fact available.
- **A late block is missed entirely.** A block first seen between the two deadlines yields no vote at
  all, though the specification permits one and the gossip rules accept it.

Both other implementations target the later deadline. Lighthouse's validator sleeps to
`duration_to_next_slot + payload_attestation_due` before fetching; Prysm's waits on the
payload-available event or that same deadline, whichever comes first. We were the outlier, and the
only client fetching at a moment its beacon node may decline to serve.

### Decision

Schedule the job at `PAYLOAD_ATTESTATION_DUE_BPS` plus a small fixed grace, and run the job's context
to the end of the slot.

- **The grace is 250ms.** A beacon node compares its own clock against the deadline before deciding
  the data is final, so firing exactly on the deadline can still be refused by a node whose clock
  trails ours. The grace covers that skew and leaves the rest of the slot for the vote to propagate.
  Landing just past the nominal deadline is also what both other clients do in practice, since each
  fetches at the deadline and only then signs and publishes.
- **The context runs to the end of the slot.** Bounding it at the deadline the job now fires at would
  cancel the signing and submission the vote depends on, and the gossip rules accept a payload
  attestation for the whole of its slot.
- **`PAYLOAD_DUE_BPS` is no longer read.** Whether the payload was timely is the beacon node's
  determination, so `payloadDueDelay` and `WithPayloadDueDelay` are removed rather than left as
  unused configuration.

### Consequences, and what stays open

A fixed deadline is a backstop rather than the best available behaviour: the vote could be cast as
soon as the beacon node reports the payload available, as Prysm's validator does. That needs the
payload-available SSE topic in `go-eth2-client`, which does not expose it yet, so a `TODO` at the
scheduling site records it.

This decision depends on `payload_present` being answered from a recorded observation timestamp
rather than from the node's view at request time. That was verified in both implementations —
Lighthouse compares `entry.timestamps.observed` against `get_payload_due()`, and Prysm reads a
`PayloadEarly(root)` flag — but it is pinned by neither `beacon-APIs`, which documents neither the
field's meaning nor when it is evaluated, nor by a released version: both were read on unstable
branches that may change before mainnet.

## Validation, before signing

`Attest` refuses to sign anything it cannot fully account for, rather than signing a best guess:

- the response must be Gloas-versioned and non-nil
- the returned data's slot must equal the duty's slot
- the number of signatures returned must equal the number of accounts submitted

A validator with no account for the duty is skipped rather than failing the whole slot's job, and a
duty with no usable accounts submits nothing at all.

## Also in this change

- Payload attestation duties are refreshed on a dependent root change, alongside the attester duty
  refreshes that already run there: the previous dependent root refreshes this epoch, the current
  dependent root refreshes the next. Each refresh cancels the epoch's existing per-slot jobs before
  rescheduling them, and leaves the current slot alone when its job has already run.
- `TestObtainPayloadAttestationTiming` (previously `TestObtainPayloadTimings`) asserted that a served
  `PAYLOAD_DUE_BPS_GLOAS` took precedence over `PAYLOAD_DUE_BPS`. The specification defines no
  `_GLOAS`-suffixed form of any payload key, because these deadlines are new in Gloas, so that case
  tested a key no chain serves. It now asserts that the unsuffixed key holds the deadline, and takes
  its expected values from the configuration `beacon.glamsterdam-devnet-8.ethpandaops.io` serves.
- `schedulePayloadAttestations` is split into `canSchedulePayloadAttestations`,
  `payloadAttestationsAlreadyScheduled` and `payloadAttestationDuties` helpers, to stay within the
  complexity limit. This is a pure extraction with no behaviour change.

### Local Devnet runtime validation — 2026-08-19

  - Built `vouch:gloas` from PR head `a18564ee03957a1a63093995b4b3a87cc4e7c719` and ran it in a fresh-genesis local Devnet-8-derived enclave.
  - During a 31-minute continuous observation interval, consensus stayed healthy and finalized; Vouch remained ready. It has now processed 35 epochs.
  - Vouch **successfully fetched PTC duties for epochs 1–36**. At the latest metric scrape, it had **successfully obtained payload-attestation data 993 times** and **made 2,217 successful per-provider payload-attestation POSTs**.
  - **PTC scheduling is active**: 1,083 payload-attestation jobs have been scheduled and 1,022 have started on their timer. Of the difference, 28 were xplicitly cancelled during dependent-root refreshes; the remaining 33 were pending when the metrics were scraped. These cumulative counters are therefore not an indication of missed duties.
  - The run exposes a Prysm-side endpoint issue: each of the three beacon nodes returned 29 failed payload-attestation-data requests (`503 no canonical shuffling block for slot`). At affected slots, all nodes still agreed on and served the canonical block/root, isolating the failure to that endpoint rather than Vouch’s duty scheduling or signing.
  - Some individual provider POSTs were cancelled, but Vouch’s multinode submitter completes when another configured beacon node accepts the message; the successful per-provider POST metrics above confirm that this path is operating.
  - Re-pulled the active Prysm beacon, Prysm validator, and Geth `glamsterdam-devnet-8` tags. All were already current; no client image digest changed.

### DIFF
--- a/services/signer/standard/service.go
+++ b/services/signer/standard/service.go
@@ -43,6 +43,7 @@ type Service struct {
 	applicationBuilderDomainType          *phase0.DomainType
 	blobSidecarDomainType                 *phase0.DomainType
 	beaconBuilderDomainType               *phase0.DomainType
+	ptcAttesterDomainType                 *phase0.DomainType
 }
 
 // Module-wide log.
@@ -115,6 +116,11 @@ func New(ctx context.Context, params ...Parameter) (*Service, error) {
 		log.Warn().Err(err).Msg("DOMAIN_BEACON_BUILDER unavailable in spec; execution payload envelope signing unavailable")
 	}
 
+	var ptcAttesterDomainType *phase0.DomainType
+	if tmp, err := domainType(spec, "DOMAIN_PTC_ATTESTER"); err == nil {
+		ptcAttesterDomainType = &tmp
+	}
+
 	s := &Service{
 		monitor:                               parameters.monitor,
 		clientMonitor:                         parameters.clientMonitor,
@@ -131,6 +137,7 @@ func New(ctx context.Context, params ...Parameter) (*Service, error) {
 		applicationBuilderDomainType:          applicationBuilderDomainType,
 		blobSidecarDomainType:                 blobSidecarDomainType,
 		beaconBuilderDomainType:               beaconBuilderDomainType,
+		ptcAttesterDomainType:                 ptcAttesterDomainType,
 	}
 
 	return s, nil
--- a/services/signer/standard/service.go
+++ b/services/signer/standard/service.go
@@ -119,6 +119,8 @@ func New(ctx context.Context, params ...Parameter) (*Service, error) {
 	var ptcAttesterDomainType *phase0.DomainType
 	if tmp, err := domainType(spec, "DOMAIN_PTC_ATTESTER"); err == nil {
 		ptcAttesterDomainType = &tmp
+	} else {
+		log.Warn().Err(err).Msg("DOMAIN_PTC_ATTESTER unavailable in spec; payload attestation signing unavailable")
 	}
 
 	s := &Service{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ gloas:
   - convert gloas aggregate attestations and signed aggregate-and-proof containers for aggregation and submission, rejecting a missing gloas arm before signing
   - derive the four controller duty deadlines from the chain specification, choosing the pre-gloas or gloas deadline per duty from that duty's slot and retaining the legacy timing for pre-gloas networks
   - default controller.max-attestation-delay, controller.attestation-aggregation-delay, controller.max-sync-committee-message-delay and controller.sync-committee-aggregation-delay to 0; the hardcoded defaults made the spec-derived deadlines unreachable
+  - add the payload timeliness committee duty: fetch PTC duties for the epoch, group each slot's validators into one job scheduled inside the gloas payload timing window, and vote on whether the slot's execution payload was revealed on time
+  - submit versioned payload attestations through the immediate, multinode and null submitters, batching signatures for validators that share a slot
   - update go-eth2-client to a gloas pseudo-version
   - satisfy the attgo struct field order and comment capitalisation rules across services and strategies
 
--- a/main.go
+++ b/main.go
@@ -63,6 +63,8 @@ import (
 	"github.com/attestantio/vouch/services/multiinstance"
 	alwaysmultiinstance "github.com/attestantio/vouch/services/multiinstance/always"
 	staticdelaymultiinstance "github.com/attestantio/vouch/services/multiinstance/staticdelay"
+	"github.com/attestantio/vouch/services/payloadattester"
+	standardpayloadattester "github.com/attestantio/vouch/services/payloadattester/standard"
 	"github.com/attestantio/vouch/services/proposalpreparer"
 	standardproposalpreparer "github.com/attestantio/vouch/services/proposalpreparer/standard"
 	"github.com/attestantio/vouch/services/scheduler"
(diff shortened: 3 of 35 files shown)


services/signer/standard/service.go
```diff
@@ -115,6 +116,11 @@ func New(ctx context.Context, params ...Parameter) (*Service, error) {
 		log.Warn().Err(err).Msg("DOMAIN_BEACON_BUILDER unavailable in spec; execution payload envelope signing unavailable")
 	}
 
+	var ptcAttesterDomainType *phase0.DomainType
+	if tmp, err := domainType(spec, "DOMAIN_PTC_ATTESTER"); err == nil {
+		ptcAttesterDomainType = &tmp
+	}
+
```
AntiD2ta: Shouldn't we add a log here like for `beaconBuilderDomainType` ?

PR #431, opened 2026-08-18
Update proposal preparations when a consensus client becomes active

### LINKED ISSUE #430: Proposal preparations are only resubmitted once per epoch, leaving a restarted beacon node with a stale fee recipient
## Summary

Proposal preparations (`POST /eth/v1/validator/prepare_beacon_proposer`) are submitted by
Vouch on a fixed once-per-epoch schedule. Beacon nodes hold that state in memory, so a beacon
node restart discards it and the node falls back to its own CLI-level default fee recipient
until Vouch's next scheduled resubmission.

If a proposal duty falls in that window **and** the block ends up being built locally by the
beacon node, the block is produced with the wrong fee recipient. This happened to us on
mainnet, and nothing in Vouch's logs or metrics at default settings indicated a problem.

Two behaviours combine to make the window larger and quieter than it first appears:

1. **The resubmission is time-based only.** `startProposalsPreparer`
   (`services/controller/standard/proposalspreparer.go`) schedules the job at
   `epochDuration*3/4 + slotDuration/2` — 294 s into each epoch on mainnet. Nothing else
   triggers a resubmission, so there is no reaction to a beacon node coming back.

2. **A submission aimed at a node Vouch currently considers inactive is silently dropped.**
   In `updateProposalPreparations`
   (`services/proposalpreparer/standard/updatepreparations.go`), `ErrNotActive` is skipped
   with a `Debug` log and is deliberately not counted as a failure, so the cycle is still
   recorded as `proposalPreparationCompleted(..., "succeeded")`:

   ```go
   if err := proposalPreparationsSubmitter.SubmitProposalPreparations(ctx, proposalPreparations); err != nil {
       if errors.Is(err, eth2client.ErrNotActive) {
           // If the client isn't ready we don't count it as a failure.
           s.log.Debug()....Msg("Client is no active; cannot update proposal preparations")
           continue
       }
       ...
   ```

   That is reasonable on its own — the node genuinely cannot accept the request — but the
   skipped submission is never retried ahead of schedule. A node that is unreachable at
   294 s into epoch N gets nothing until 294 s into epoch N+1, however long before that it
   came back.

The net effect is that the restarted node can be up, answering, and synced for several
minutes while holding no fee-recipient preferences at all.

## Why the state is lost on the beacon node side

`prepare_beacon_proposer` is documented as needing periodic resubmission, and at least Prysm
keeps it purely in memory — its handler writes to a cache with no persistence
(`beacon-chain/rpc/eth/validator/handlers.go`, `PrepareBeaconProposer`):

```go
s.ProposerPreferencesCache.SetDefault(cache.ProposerPreference{
    ValidatorIndex: primitives.ValidatorIndex(validatorIndex),
    FeeRecipient:   bytesutil.ToBytes20(feeRecipient),
})
```

So the preferences do not survive a restart, and until Vouch resubmits, the node uses its own
`--suggested-fee-recipient` for any block it builds.

## Observed timeline (mainnet)

Vouch 1.13.1, multiple beacon nodes, `blockrelay` with a `proposerConfig` assigning
per-validator fee recipients. One beacon node (call it **node B**) was rebooted by an
unattended OS upgrade. Times are relative to the moment Vouch lost its connection to node B;
validator index, slot numbers and hostnames are redacted.

| Time | Event |
|---|---|
| T+0 s | Node B goes down. Vouch's event-stream subscription drops. |
| T+41 s … T+162 s | Vouch reports node B as `client is not active` on attestation/aggregate submissions. |
| **T+91 s** | **Scheduled preparation cycle for the current epoch runs.** Node B is marked inactive, so its submission is skipped via `ErrNotActive`. Not logged at default level; cycle recorded as succeeded. |
| T+174 s | Node B is answering again (errors change from `not active` to `not synced`) — i.e. back up, ~1.5 min after the skipped cycle. |
| T+210 s | Last `not synced` line; node B fully back. |
| **T+475 s** | Next scheduled preparation cycle runs — the first one since node B returned, ~5 minutes after it came back. |
| **T+481 s** | Start of our proposal slot, ~6 s later. |

For that proposal the relay unblinding failed, so the block was built locally by node B — and
it carried node B's CLI-level default fee recipient rather than the per-validator address from
`proposerConfig`.

The ~6 s gap between the resubmission and the slot is not a safe margin in practice: a beacon
node begins building the payload for slot *N* before slot *N* starts, so the fee recipient in
use for that block was fixed before the resubmission landed. Effectively there were no valid
preparations on node B for the entire ~6.5 minutes between its restart and the proposal.

## Redacted log extract

Connection to node B drops (T+0):

```json
{"level":"error","service":"client","impl":"http","id":"<redacted>","address":"http://<node-b>:5051","error":"unexpected EOF","time":"T+0.000","message":"Failed to subscribe to event stream"}
```

Node B marked inactive — this is the state in which the scheduled preparation cycle at T+91 s
was skipped (T+41 s onward):

```json
{"level":"warn","strategy":"submitter","impl":"multinode","beacon_node_address":"<node-b>:5051","slot":<redacted>,"error":"client is not active","time":"T+41.474","message":"Failed to submit aggregate attestations"}
{"level":"warn","strategy":"submitter","impl":"multinode","beacon_node_address":"<node-b>:5051","slot":<redacted>,"error":"client is not active","time":"T+42.497","message":"Failed to submit attestations"}
```

Node B answering again ~1.5 minutes later (T+174 s), still no preparation resubmission until
T+475 s:

```json
{"level":"warn","strategy":"submitter","impl":"multinode","beacon_node_address":"<node-b>:5051","slot":<redacted>,"error":"client is not synced","time":"T+173.957","message":"Failed to submit attestations"}
```

There are **no** proposal-preparation log lines anywhere in the window — the skip is `Debug`
and success is `Trace`, so at default log level the whole sequence is invisible.

## Impact

- Any proposal duty falling between a beacon node restart and Vouch's next scheduled
  preparation cycle can be built with the wrong fee recipient, if that node builds the block
  locally (relay unavailable, unblinding failure, `min_value` not met, builder disabled, …).
- The exposure window is up to one full epoch (~6.4 min on mainnet) after a restart, and up to
  two if the restart straddles a scheduled cycle, as above — the node can be fully back and
  synced for most of that time.
- The proceeds go to the beacon node's default fee recipient, which is generally not
  recoverable.
- It is silent: no error or warning at default log level, and
  `vouch_proposalpreparation_process_requests_total` still increments `result="succeeded"`
  for a cycle in which every node was skipped.

I would expect this to be more likely to bite operators who run unattended OS upgrades on
their beacon nodes, and less visible for those whose blocks almost always come from a relay,
since the wrong fee recipient only materialises on a locally-built block.

## Suggestion

Rather than only resubmitting on the epoch schedule, also resubmit proposal preparations when
a beacon node transitions back to active — i.e. push `prepare_beacon_proposer` on reconnect,
in addition to the existing per-epoch cycle. `go-eth2-client`'s HTTP service already exposes
an `OnActive` hook (`http/hooks.go`, `WithHooks`) fired from its connection-state check, so
the transition is already detected; Vouch does not currently pass hooks when constructing its
clients.

Happy to put together a PR if that direction sounds right.

## Environment

- Vouch 1.13.1
- go-eth2-client v0.29.0
- Mainnet, multiple beacon nodes configured, Dirk account manager
- `blockrelay` with a `proposerConfig` defining per-validator fee recipients

ThomasDalla:
Fixes #430.

## What this changes

Beacon nodes hold proposal preparations in memory, so a node that restarts has none until
Vouch's next scheduled update. That update runs once per epoch, and a node that is inactive
when it runs is skipped via `ErrNotActive` and not retried until the following epoch — so a
node can be back up and synced for minutes with no fee-recipient preferences. A proposal in
that window, if the block is built locally, uses the beacon node's own default fee recipient.

This updates proposal preparations as soon as a consensus client becomes active, in addition
to the existing per-epoch update.

## Why this shape

- **`go-eth2-client` already detects the transition.** `http.Hooks.OnActive` is fired from
  `CheckConnectionState` when a client goes from inactive to active, on the existing 30-second
  connection-state poll. Vouch simply never passed hooks when constructing its clients, so the
  change is to provide them — no new polling, timers or connection tracking.

- **Callbacks are registered in `clients.go`** because `fetchClient` is where every client is
  constructed and cached, and the hooks have to be supplied at construction time. The hook
  reads the callback list when it fires rather than capturing it, so it does not matter whether
  a client is constructed before or after a callback is registered.

- **`UpdatePreparations` is called rather than targeting the client that just became active.**
  It already submits to every configured submitter in one call, which is cheap, and it means a
  reconnect that coincides with another node also being unavailable still ends up updating both
  once they are back. Targeting a single node would need address-matching against the submitter
  list for no practical gain.

- **The `nil` check on `proposalPreparer` is deliberate.** `initProposalPreparer` returns a nil
  service (with a nil error) when the chain is not yet bellatrix-capable; without the check the
  first activation would panic inside a client-owned goroutine.

- **Callbacks are called synchronously.** `OnActive` is already invoked in its own goroutine by
  the client, so there is nothing to gain from another one, and this keeps a panic attributable.
  The callback list is copied under the read lock before dispatch, so a callback that registers
  another callback cannot deadlock — covered by a test.

## What this does not cover

- Connection state is polled every 30 seconds, so a beacon node that restarts and starts
  answering again within a single poll interval is never observed as inactive and no update is
  triggered. This makes the exposure window small in the common case rather than removing it
  entirely. Tying preparation updates to known proposal duties would close it properly, but
  that is a larger change and I did not want to bundle it here.

- Validator registrations (`blockrelay`) have the same once-per-epoch shape and could reuse
  the same callback, but they were not the cause of the issue, so I have left them alone to
  keep this focused. Happy to add that here or separately if you would like it.

## Testing

- `go build ./...`, `go vet ./...` and `go test ./...` all pass.
- New unit tests cover callback dispatch, the no-callbacks case, dispatch on each activation,
  and registration from within a callback; they pass under `-race`.
- I intend to run this build against mainnet beacon nodes and can report back on how it
  behaves after a real beacon node restart.

Happy to adjust naming, comments, or add a CHANGELOG entry if you would like this in a
different form.

### DIFF
--- a/clients.go
+++ b/clients.go
@@ -33,6 +33,43 @@ var (
 	knownClientsMu sync.Mutex
 )
 
+var (
+	reconnectCallbacks   []func(ctx context.Context, address string)
+	reconnectCallbacksMu sync.RWMutex
+)
+
+// addReconnectCallback adds a function that is called whenever a consensus client becomes
+// active, either on first connection or on reconnection after the client has been
+// unavailable.  The client dispatches its hooks in a goroutine of their own, so callbacks
+// never run on a request path; they are called one after another, so a slow callback delays
+// those registered after it.
+func addReconnectCallback(callback func(ctx context.Context, address string)) {
+	reconnectCallbacksMu.Lock()
+	defer reconnectCallbacksMu.Unlock()
+
+	reconnectCallbacks = append(reconnectCallbacks, callback)
+}
+
+// onClientActive calls the registered reconnect callbacks for the client at the given address.
+func onClientActive(ctx context.Context, address string) {
+	reconnectCallbacksMu.RLock()
+	callbacks := make([]func(context.Context, string), len(reconnectCallbacks))
+	copy(callbacks, reconnectCallbacks)
+	reconnectCallbacksMu.RUnlock()
+
+	for _, callback := range callbacks {
+		callback(ctx, address)
+	}
+}
+
+// clientHooks are the hooks provided to each consensus client, allowing vouch to react to
+// changes in the client's connection state.
+var clientHooks = &httpclient.Hooks{
+	OnActive: func(ctx context.Context, s *httpclient.Service) {
+		onClientActive(ctx, s.Address())
+	},
+}
+
 // fetchClient fetches a client service, instantiating it if required.
 func fetchClient(ctx context.Context, monitor metrics.Service, address string) (eth2client.Service, error) {
 	if address == "" {
@@ -55,6 +92,7 @@ func fetchClient(ctx context.Context, monitor metrics.Service, address string) (
 				"User-Agent": fmt.Sprintf("Vouch/%s", ReleaseVersion),
 			}),
 			httpclient.WithReducedMemoryUsage(util.HierarchicalBool("reduced-memory-usage", fmt.Sprintf("eth2client.%s", address))),
+			httpclient.WithHooks(clientHooks),
 		)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to initiate consensus client")
--- a/clients_test.go
+++ b/clients_test.go
@@ -1,0 +1,157 @@
+// Copyright © 2026 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"testing"
+
+	mockproposalpreparer "github.com/attestantio/vouch/services/proposalpreparer/mock"
+	"github.com/stretchr/testify/require"
+)
+
+// resetReconnectCallbacks clears the registered callbacks, restoring them when the test ends.
+func resetReconnectCallbacks(t *testing.T) {
+	t.Helper()
+
+	reconnectCallbacksMu.Lock()
+	existing := reconnectCallbacks
+	reconnectCallbacks = nil
+	reconnectCallbacksMu.Unlock()
+
+	t.Cleanup(func() {
+		reconnectCallbacksMu.Lock()
+		reconnectCallbacks = existing
+		reconnectCallbacksMu.Unlock()
+	})
+}
+
+func TestOnClientActiveWithoutCallbacks(t *testing.T) {
+	resetReconnectCallbacks(t)
+
+	require.NotPanics(t, func() {
+		onClientActive(context.Background(), "http://localhost:5051")
+	})
+}
+
+func TestOnClientActiveCallsCallbacksWithAddress(t *testing.T) {
+	resetReconnectCallbacks(t)
+
+	addresses := make([]string, 0)
+	addReconnectCallback(func(_ context.Context, address string) {
+		addresses = append(addresses, address)
+	})
+	addReconnectCallback(func(_ context.Context, address string) {
+		addresses = append(addresses, address)
+	})
+
+	onClientActive(context.Background(), "http://localhost:5051")
+
+	require.Equal(t, []string{"http://localhost:5051", "http://localhost:5051"}, addresses)
+}
+
+func TestOnClientActiveCallsCallbacksOnEachActivation(t *testing.T) {
+	resetReconnectCallbacks(t)
+
+	calls := 0
+	addReconnectCallback(func(_ context.Context, _ string) {
+		calls++
+	})
+
+	onClientActive(context.Background(), "http://localhost:5051")
+	onClientActive(context.Background(), "http://localhost:5052")
+
+	require.Equal(t, 2, calls)
+}
+
+// TestRegisterProposalPreparationsUpdaterWithoutPreparer confirms that a nil preparer, as
+// provided before the bellatrix fork, does not register a callback that would panic when a
+// client becomes active.
+func TestRegisterProposalPreparationsUpdaterWithoutPreparer(t *testing.T) {
+	resetReconnectCallbacks(t)
+
+	registerProposalPreparationsUpdater(context.Background(), nil)
+
+	reconnectCallbacksMu.RLock()
+	registered := len(reconnectCallbacks)
+	reconnectCallbacksMu.RUnlock()
+	require.Equal(t, 0, registered)
+
+	require.NotPanics(t, func() {
+		onClientActive(context.Background(), "http://localhost:5051")
+	})
+}
+
+// ctxCapturingProposalPreparer records the context with which it was last called.
+type ctxCapturingProposalPreparer struct {
+	ctx context.Context
+}
+
+func (s *ctxCapturingProposalPreparer) UpdatePreparations(ctx context.Context) error {
+	s.ctx = ctx
+
+	return nil
+}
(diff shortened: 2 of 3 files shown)


ThomasDalla:
Tested this against mainnet beacon nodes, and it turned up a bug in my own first version — now fixed, and the branch has been force-pushed.

**Setup:** three configured beacon nodes, one of them restarted deliberately. Timings below are relative to the moment the node's connection dropped.

The hook fired as intended: **T+31s** after the drop, the node was active again and proposal preparations were submitted to all three nodes — nowhere near the scheduled cycle (the previous one had run ~40s before the restart, the next was ~5 minutes out).

But all three submissions failed:

```
{"level":"error","service":"proposalpreparer","impl":"standard","client":"http://<node-a>:5051",
 "error":"failed to send proposal preparations\nfailed to call POST endpoint\nPost \"http://<node-a>:5051/eth/v1/validator/prepare_beacon_proposer\": context canceled",
 "message":"Failed to update proposal preparations"}
```

**Cause:** I was using the context supplied to the hook. `assertIsActive` calls `CheckConnectionState(ctx)` on the way in to a request that finds the client inactive, so `OnActive` can be invoked with that request's context — which is cancelled the moment the request completes, cancelling the preparation POSTs in flight. Here the trigger was an attestation submission against the recovering node, so the context was gone almost immediately.

Worth noting the failure mode is quiet in the wrong direction: the reconnect update *appears* to run, and `vouch_proposalpreparation_process_requests_total{result="failed"}` increments, but no preparations actually land.

**Fix:** the callback now uses the service context (`startServices`'s, which lives for the process) and ignores the hook's. Added a regression test that fires the callback with an already-cancelled context and asserts the preparer receives a live one — it fails against the previous version with exactly `context canceled`.

If you would rather the hook contract were "the context is request-scoped, don't retain it", I am happy to move the `context.WithoutCancel`-style handling into `go-eth2-client` instead, or to document it there — just say which you prefer.

I will re-run the same restart test against the fixed build and report back.

ThomasDalla:
Confirmed working against mainnet. Three configured beacon nodes, one of them restarted deliberately; timings are relative to the moment its connection dropped:

| T+ | Event |
|---|---|
| 0s | Node goes down; event-stream subscription drops |
| 0s – 2m39s | `client is not active` on submissions to that node |
| **~2m45s** | Node active again → **proposal preparations submitted successfully to all configured nodes** |
| 2m51s | `client is not synced` — the update landed while the node was still syncing, before it could be asked to propose |

`vouch_proposalpreparation_process_requests_total{result="succeeded"}` incremented at that point, with no `result="failed"` series and no `proposalpreparer` error lines. The next scheduled update was ~3 minutes later, so the increment is unambiguously the reconnect path rather than the periodic one — without this change the node would have held no preparations for that entire window.

Duties were unaffected throughout: submissions to the restarted node failed with `client is not active` for the duration of the outage, but the multinode submitter covered them via the other two nodes and `vouch_attestation_process_requests_total{result="succeeded"}` kept climbing with no failures.

Two things worth flagging from the run:

- **The update also fires once shortly after startup**, as clients first transition to active. Harmless and arguably useful, but easy to suppress if you would rather it only ran on genuine reconnections.
- **The callback deliberately uses the service context, not the one supplied to the hook.** `assertIsActive` checks connection state on the way in to a request that finds the client inactive, so `OnActive` can be invoked with that request's context, which is cancelled as soon as the request completes — long enough to start the preparation POSTs and not long enough to finish them. If you would prefer that handled in `go-eth2-client` instead (passing a non-cancellable context to the hooks, or documenting that the hook's context is not for retained work), I am happy to do it there.

The node here took ~2m45s to become active again, well past the 30s connection-state poll, so this run does not exercise the limitation noted in the description — a restart fast enough to go unobserved between polls still gets no update.

Node identifiers and validator details omitted; happy to provide anything more specific that would help review.

PR #441, opened 2026-09-12
Add Gloas proposal auction observability

AntiD2ta:
Observe the Gloas proposal auction from request through publication without changing selection or publication behavior.

- records a bounded selection counter by strategy, requested preference class, and self-build, P2P-builder, or builder-API source
- emits structured per-provider completion, rejection, error, cancellation, and deadline outcomes without exposing configured endpoint details
- records the final selected provider, source, value state, requested policy, proposal root, and block-only or block-and-envelope publication path
- correlates request, provider, selection, block-publication, and envelope-publication spans
- keeps unknown execution values and all-unknown fallback selections explicit

Verification:
- `gosilent test ./...` (1116 tests)
- `go build ./...`
- custom lint reports only findings already present below this stack layer

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


AntiD2ta:
### `StableProviderName` does not hide the endpoint it hashes

The doc comment says "returns a stable telemetry name without exposing a configured endpoint", and the PR description says outcomes are emitted "without exposing configured endpoint details". The hash does not achieve that, because the preimage space is tiny and guessable rather than secret:

```go
digest := sha256.Sum256([]byte(provider))
return "provider-" + hex.EncodeToString(digest[:8])
```

It is an unsalted SHA-256 of the configured address string. Anyone holding a `provider-<hex>` value and a guess at the host can recover the address by enumeration. Reproduction, against a hypothetical deployment:

```python
import hashlib
host = "beacon-01.example.com"
table = {hashlib.sha256(f"{host}:{p}".encode()).hexdigest()[:16]: p
         for p in range(1, 65536)}
# every port on one host: ~0.3s on a laptop, no optimisation
```

One host's entire port range inverts in well under a second. In practice the search is far smaller, since consensus clients sit on a handful of conventional ports. I confirmed this against a real deployment before filing: all four provider ids in our logs inverted to their exact `host:port` immediately, which recovers both the endpoint list and which client produced every per-provider outcome in the telemetry.

Worth being explicit that **truncation is not the weakness**. Going from `digest[:8]` to the full 32 bytes changes nothing, because the attack is enumeration of a small input domain, not a collision or a brute-force of the digest. Anything deterministic and unkeyed over `host:port` has this property.

This reads as a real gap rather than a nitpick given the surrounding code already redacts endpoints out of error strings — the intent elsewhere in the file is clearly that endpoints should not reach telemetry.

### Suggested fixes

Two options, depending on whether the name needs to be stable across restarts.

**1. Index by configuration order** — simplest, and keeps stability:

```go
// assigned once at construction from the configured provider list
return fmt.Sprintf("provider-%d", idx)
```

No preimage to recover. It leaks only the ordering of the configured list, which is not sensitive, and it stays stable across restarts and across processes sharing a config.

**2. Keyed hash with a configured secret** — if an opaque, non-positional name is preferred:

```go
mac := hmac.New(sha256.New, s.telemetryKey)
mac.Write([]byte(provider))
return "provider-" + hex.EncodeToString(mac.Sum(nil)[:8])
```

Not invertible without the key.

One caveat against a tempting third option: **do not use a per-process random salt.** It would close the leak, but it makes the names unstable across restarts, and that stability has real operational value. We used it this week to follow a single provider's behaviour across a Vouch restart while diagnosing a proposal issue — correlating outcomes either side of the restart is only possible because the name is stable. A per-process salt would have made that impossible. Option 1 or 2 keeps the property the type name promises.

Happy to send a patch for whichever direction you prefer.

AntiD2ta:
Addressed [the provider-name privacy report](https://github.com/attestantio/vouch/pull/441#issuecomment-5796503594) in the commit titled **Redact configured beacon endpoints in Vouch telemetry**. Configured beacon providers now use stable configuration indices across Vouch telemetry; client details remain separate, optional fields.

